PR #185, opened 2026-07-20
fix(mobile-admin): repair tenant_id GIP claim + stop the login bounce loop

Sam123ben:
## Problem

Mobile admin login bounced endlessly back to the sign-in screen — for **every tenant created through the onboarding wizard**, on **every provider** (password, Google, Apple). Web was unaffected.

Two independent defects compounded:

**1. Nothing ever wrote the `tenant_id` custom claim.**
`marketplace-api` resolves a mobile caller's tenant *solely* from the `tenant_id` claim on the GIP ID token (`gip_verifier.go:34`). But there is **not a single `SetCustomUserClaims` call anywhere in the repo** — `gipadmin` only did password resets. Onboarding created the tenant, store, and FGA tuples, and never touched GIP claims.

Production state when diagnosed — **4 of 5 tenants locked out**:

| Tenant | claim |
|---|---|
| The Bondi Store | `{"tenant_id":["8c302556…"]}` (set by hand) |
| The Facade Factory | none |
| My God | `{"admin":true}` — no tenant_id |
| Apple Store | none |
| Phone Store | none |

**2. A missing claim was reported as 401.**
`gip_verifier` returned an error → `GIPBearerAuth` → `401 invalid or expired token` → mobile client reads a 401 surviving refresh as a dead session → `signOut()` → router redirects to `/login`. An oscillation with no error shown.

Web is unaffected because it resolves the tenant from the **database**, never the claim.

## Fix

**`marketplace-api`** — "no store" is an *authorization* fact, not an authentication failure. Pass through with an empty `tenant_id` and let `authz.RequireTenantRelation` answer 404, which `TenantGate` already renders as **"No store yet"**. That empty state existed all along and could never render, because the 401 tore the session down first. Invalid tokens still 401.

**`platform-api`**
- `gipadmin.EnsureTenantClaim` — writes `tenant_id` as an array; **merges** rather than clobbers. GIP replaces `customAttributes` wholesale and a real account carries `{"admin":true}`, so a naive write would silently strip it. Idempotent → outbox redelivery is safe.
- Enqueued **in the same transaction as the FGA tuple**, via the outbox — deliberately *not* best-effort like `EnsureSelfVendor`, because a silent failure locks the owner out of mobile entirely. Retries with backoff.
- `cmd/backfill-gip-claims` — repairs pre-existing owners. Idempotent, `-dry-run`, continues past individual failures.

## Tests

New unit tests for: 401-vs-404 contract (both directions), claim merge preserving unrelated claims, multi-tenant append, outbox handler error propagation (so the drainer retries), payload validation.

A test caught a real bug during development: `existingTenantIDs` handled `string` and `[]any` but not `[]string`, so appending a **second** tenant silently dropped the first. Fixed.

> [!WARNING]
> Integration tests compile and vet clean but were **not executed** — no local Docker, and CI skips them without `TEST_DATABASE_URL`. Coverage here is unit tests plus direct verification against prod GIP/DB state.

## Post-merge steps (required)

1. Run `cmd/backfill-gip-claims` against prod for the 4 affected owners
2. **Owners must sign out and back in** — claims only refresh on a new ID token

## Out of scope

- **Web "Sign in with Apple"** cannot be added yet: the Apple provider is configured native-only (`bundleIds: ["com.mark8ly.admin"]`, `clientId`/`clientSecret` empty). Web needs an Apple **Services ID**, a `.p8`-signed client secret (**expires every 6 months**), a registered return URL, and domain verification — all Apple Developer Portal actions.
- **Apple "Hide My Email"** still forks a separate account by design; the existing Settings → Security link flow is the supported path.

## Design note

The claim is a **snapshot** — it goes stale if membership changes and only refreshes on a new ID token. Resolving tenancy server-side from the DB (as web does) would be more robust. This PR makes the existing claim-based path correct rather than replacing it.

### DIFF
--- a/services/marketplace-api/internal/auth/gip_verifier.go
+++ b/services/marketplace-api/internal/auth/gip_verifier.go
@@ -29,10 +29,17 @@ func (v *GIPVerifier) Verify(ctx context.Context, idToken string) (*TokenClaims,
 		return nil, ErrInvalidToken
 	}
 
+	// A missing tenant_id claim is NOT an authentication failure — the
+	// token is validly signed and the user is who they say they are.
+	// They simply have no store bound to this identity yet.
+	//
+	// Returning an error here made GIPBearerAuth abort with 401, which
+	// the mobile client reads as "session is dead" and responds to by
+	// calling signOut() — bouncing the user back to /login with no
+	// explanation, in a loop. Instead, pass through with an empty
+	// TenantID and let authz.RequireTenantRelation answer 404, which
+	// the app already renders as its "No store yet" screen.
 	tenantID := tenantIDFromClaims(token.Claims)
-	if tenantID == "" {
-		return nil, fmt.Errorf("token missing tenant_id claim")
-	}
 
 	return &TokenClaims{
 		UserID:   userID,
--- a/services/marketplace-api/internal/auth/gip_verifier_test.go
+++ b/services/marketplace-api/internal/auth/gip_verifier_test.go
@@ -1,6 +1,14 @@
 package auth
 
-import "testing"
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/gin-gonic/gin"
+)
 
 // tenant-service writes tenant_id as a multi-value array (its
 // UpdateUserAttributes appends into []interface{}), so the array form is what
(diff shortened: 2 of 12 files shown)


Sam123ben:
## Security follow-up — fail-open closed

The automated security review flagged the verifier change as a potential **fail-open**. It was right, and `a665271` fixes it.

Admitting a claimless token removed the 401 that was *implicitly* protecting routes without an explicit authz guard. An audit of every route under `bearerAuth` found three:

| Route | Previous protection |
|---|---|
| `/mobile/admin/platform-support` | **none** — and `platformResolver` forwards `c.GetString("tenant_id")` to otto as `X-Client-Tenant-Id` |
| `push-tokens` POST | `StoresMiddleware` only |
| `push-tokens` DELETE | `StoresMiddleware` only |

Concretely: an authenticated user with no store could have opened platform-support chats attributed to an empty tenant. Not a cross-tenant data read — otto's scope is pinned to the platform tenant — but a genuine broadening of access, and the reviewer's label is fair.

**Fix:** `auth.RequireTenantClaim()` aborts with **404** — matching `RequireTenantRelation`'s no-existence-leak convention, and deliberately **not 401**, since 401 is precisely what makes the mobile client `signOut()` and bounce. Mounted alongside `bearerAuth` at **group level**, so routes added later fail closed by default rather than depending on someone remembering a check.

Three tests added: blocks empty tenant, blocks a missing key entirely, allows a bound tenant — plus an explicit assertion that the guard never returns 401.

Issue #366, opened 2026-08-25
Production Stripe billing key is test mode — record the go-live transition before it is swapped


mahesh-sangawar:
Found while scoping #358, which describes pushing `trial_end` to Stripe as "a real-money external side effect". It is not one today, and that is worth recording before it becomes one.

## Measured 2026-08-26

GCP Secret Manager is the source of truth (External Secrets syncs it into the `mark8ly` namespace):

- `prod-mark8ly-stripe-billing-secret-key` → `sk_test_…`
- `prod-mark8ly-uat-stripe-billing-secret-key` → `sk_test_…`
- k8s secret `mark8ly-stripe-billing` in namespace `mark8ly` → `sk_test_…` (so the cluster copy is not stale; it agrees)

Only the first eight characters were read. The key itself is not reproduced here or anywhere else.

**Production billing runs against Stripe test mode.** This is consistent with `store_subscriptions` holding 0 rows — no merchant has entered the billing flow, so nothing has been charged and nothing could have been.

## Why it matters

Every Stripe write path in `marketplace-api` is currently exercised against a sandbox. The moment this secret is swapped to a live key, the following start moving real billing dates and real money, with no code change and no deploy to mark the transition:

- `internal/billing/trial/subscribe.go` — `CreateSubscription` with `trial_end`
- `internal/subscription/planchange` — `UpdateSubscription` price swaps (both callers)
- #358's trial extension, which additionally moves `billing_cycle_anchor`
- the hard-delete runner's Stripe customer deletion

## What this issue asks for

Not a fix — a decision recorded before the swap:

1. A deliberate go-live checklist for the key swap, naming the paths above.
2. Whether anything should assert key mode at startup (e.g. log loudly, or refuse a `sk_live_` key outside a specific env), so "we are on live now" is observable rather than inferred.
3. Whether uat and prod should ever hold the same key value, as they do today.

## Verification note

#358's acceptance requires exercising the extension against a real Stripe subscription, which production's 0 rows cannot provide. Because the configured key is test mode, that verification can use the estate's own key without any live-money exposure. That convenience disappears at the swap.

mahesh-sangawar:
Execution issue filed as #371 — the swap itself, including two pre-requisites that are easy to miss: the webhook secret (`prod-mark8ly-stripe-billing-webhook-secret`) has its own live-mode value, and the live account needs its own Price objects with byte-identical lookup keys or every subscribe and plan change fails at price resolution.

[...]

Issue #336, opened 2026-08-24
outbox publisher marks dropped events as published, and outbox_events.error is never written


mahesh-sangawar:
Found while building `/admin/health` (#289), which originally specified an outbox `errored` metric. The metric had to be dropped: it could never have returned a non-zero value. Investigating why turned up a second, worse problem.

## 1. `outbox_events.error` is never written by anything

The column exists (`migrations/000001_products_initial.up.sql:256`) and is declared on the model (`internal/outbox/models.go`). Searching every write to that table across `services/marketplace-api` finds exactly one:

```go
// internal/outbox/repository.go:65
UPDATE outbox_events SET published_at = now() WHERE id IN ?
```

Nothing sets `error`. Ever. The publisher logs failures and moves on, so the reason an event failed is never persisted.

**This blocks #331.** That issue defines its `failed` status as `published_at IS NULL AND error IS NOT NULL` and describes the value as "a row with `published_at IS NULL` and a non-null `error`, hours old". That filter can never match a production row, so `status=failed` would be permanently empty while looking like it worked. Worth settling this before #331 is built, not after.

## 2. Worse: dropped events are marked as successfully published

`Publisher.Tick` (`internal/outbox/publisher.go:85`) appends every row's id to `ids` **before** its two validation checks:

```go
for _, r := range rows {
    ids = append(ids, r.ID)          // <-- appended first, unconditionally
    if err := json.Unmarshal(r.Payload, &payload); err != nil {
        logger.Warn("outbox publisher: unparseable payload; dropping", ...)
        continue                      // <-- skips the watermark, keeps the id
    }
    sid, _ := payload["store_id"].(string)
    if sid == "" {
        logger.Warn("outbox publisher: payload missing store_id; dropping", ...)
        continue                      // <-- same
    }
    ...
}
...
return p.repo.MarkPublishedInTx(tx, ids)  // marks the dropped ones too
```

So an event with an unparseable payload, or a payload missing `store_id`, is:

- **dropped** — its `store_watermarks` row is never bumped, so the downstream consumer never learns the aggregate changed, and
- **marked `published_at = now()`** anyway.

The event ends up indistinguishable from a successful publish. It is not `pending`, not `failed`, and invisible to any query. The only trace is a `Warn` line in the pod log, which nothing aggregates.

[...]

Issue #317, opened 2026-08-23
Integration fixtures: store FK and stores NOT NULL drift across 17 packages


mahesh-sangawar:
Found by the full `-tags=integration ./...` run for marketplace-api. Companion to #316. Context and the production bug the dark suite was hiding: #315.

Two clusters, same root cause as #316 — migrations added constraints, fixtures never followed. **Both were already fixed once in #315**, in `dunning` and `tenantpurge` respectively, so the fix pattern is established and copyable.

## Cluster A — `store_subscriptions` store FK (13 packages)

Tests insert `store_subscriptions` for a random `store_id` with no `stores` row, violating `store_subscriptions_store_id_fkey`. There is no seed binary for this service and `stores` is empty locally, so nothing else provides it.

Representative sites:
- `internal/arbitrage/appeal_test.go:40`
- `internal/subscription/statemachine/cas_conflict_test.go:34`

**Fix:** reuse the helper added in #315 at `internal/subscription/dunning/testhelpers_integration_test.go`. Note its deliberate difference from `internal/order`'s version — the caller supplies the tenant, so the store and subscription agree about tenancy. Without that, tenant-scoped assertions pass while testing nothing. Consider promoting it to a shared test-support package rather than copying it 13 times.

## Cluster B — `stores.storefront_customer_portal_secret` NOT NULL (4 packages)

Tests insert `stores` rows omitting a `NOT NULL` column added by a later migration.

Sites:
- `internal/category/repository_integration_test.go:64`
- `internal/handlers/internalsvc/storefront_status_test.go:81`
- `internal/outbox/publisher_integration_test.go:58`
- `internal/subscription/planchange/criterion_39_test.go:41`

**Fix:** as in #315's tenantpurge commit — `encode(gen_random_bytes(32), 'hex')` in the INSERT.

## Smaller items in the same sweep

- **`internal/apikeys`** — a `varchar(60)` overflow: `lastused_test.go:33`, `repo_test.go:37,59,77`
- **`internal/orderrefund`** — deadlocks in shared testdb cleanup (`testdb.go:68`) plus a config-wiring gap (`resolver_integration_test.go:239`)
- **`internal/page/repository_integration_test.go:60`** — reports "transaction is aborted", which is a cascade symptom; the real cause was not visible in the excerpt and needs its own look

## Verify

```
cd services/marketplace-api
TEST_DATABASE_URL='postgres://dev:dev@localhost:5432/marketplace_db?sslmode=disable' \
[...]

Issue #282, opened 2026-08-22
Platform integration: GET /admin/kpis — headline counters, 501 when uninstrumented


mahesh-sangawar:
Part of the platform console integration series (see #274). Depends on #275.

### Build

`GET /admin/kpis` — a flat map of headline business metrics for the console's product tile: active tenants, stores, GMV, onboarding in flight, trials expiring.

```json
{ "tenants_active": 412, "stores_active": 501, "onboarding_in_flight": 12 }
```

Keys are product-defined; the console renders label and format from its own registry.

### Do not return `{}` for anything uninstrumented

Return `501 not_implemented` — for the key or for the endpoint.

The console's existing KPI route falls through to an empty object for products it does not recognise, which is why one product has rendered four em-dashes since launch: **dashes that look like zeroes**. "Not instrumented" and "zero" are different answers and must stay distinguishable.

### Money

Minor units, always with an explicit currency:

```json
{ "gmv_today": { "amount": 984200, "currency": "INR" } }
```

Never a bare number. Three different money conventions across one console is a live footgun this contract exists to close.

### Acceptance

- [ ] Uninstrumented metrics yield `501`, never `{}` and never `0`
- [ ] Every money value carries a currency and is an integer in minor units

mahesh-sangawar:
Delivered in #325 and **verified in production** on `main-1eb5631` (both `mark8ly-platform-api` and `mark8ly-marketplace-api`, promoted through Kargo).

Path: `https://api.mark8ly.com/api/v1/platform/admin/kpis`

## Live results

```
GET /admin/kpis
{"data":{"onboarding_in_flight":0,"stores_active":4,"tenants_active":4,"trials_expiring":0}}
```

Both acceptance criteria hold against the live deployment:

| case | live response |
|---|---|
| `?keys=gmv_today` | `501` — `kpi "gmv_today" is known but not instrumented` |
| `?keys=not_a_real_kpi` | `501` — `kpi "not_a_real_kpi" is not a recognised key` |
| `?keys=tenants_active,gmv_today` | `501` naming `gmv_today` — **not** a partial `200` |

Every `501` carries a machine-readable `key` field, so the console can mark one tile rather than the whole panel. **No uninstrumented metric returns `{}`, and none returns `0`.**

The two messages genuinely differ in production, not just in tests — that was worth confirming on the wire, because the status code is identical and the message is the only thing separating "we know this metric and haven't built it" from "that is not one of our keys".

[...]

PR #367, opened 2026-08-25
feat(billing): push a changed trial_end to Stripe for card-backed trials (#358)

### LINKED ISSUE #358: Trial extension: push a changed trial_end to Stripe for card-backed trials
Split out of #286, which refuses these with `409 stripe_managed`.

## The situation

`POST /admin/billing/trials/{store_id}/extend` refuses any subscription with a `stripe_subscription_id`. That covers the common support case — a merchant who has not added a card needs more time — but leaves no path for a trial where the merchant HAS added a card and Stripe now holds the billing date.

Extending such a trial locally without telling Stripe would put the console and Stripe in disagreement about when a real merchant is charged. That is the split-brain #353 existed to remove, so #286 refuses rather than half-does it.

## What it actually takes

**Not a new client method.** `internal/billing/stripe/update.go:37` already has `UpdateSubscription`. The blocker is narrower:

- `UpdateSubscriptionParams` carries `SubscriptionID`, `PriceID`, `ProrationBehavior`, `IdempotencyKey`, `Metadata` — and **no `TrialEnd`**.
- `UpdateSubscription` *requires* a `PriceID` and swaps the subscription's item (`update.go:47-56`); it exists for plan changes. Calling it to extend a trial would **silently re-price the subscription**.

So the work is: add `TrialEnd` to the params, make the price swap optional, and only then wire it behind the refusal. `sdk.SubscriptionUpdateParams` supports `TrialEnd` natively, so it is small — but it is a **real-money external side effect** and needs its own failure semantics.

## The decision this issue has to make

What happens when the local write succeeds and the Stripe call fails? The options are not equivalent:

- **Refuse up-front, call Stripe first, then write locally** — Stripe becomes the source of truth for card-backed trials. A local write failure then leaves Stripe ahead, which is the safer direction (the merchant is charged later than the console shows, not earlier).
- **Write locally, then call Stripe, roll back on failure** — needs a compensating write, and a crash between the two leaves them disagreeing anyway.
- **Write locally, enqueue the Stripe update** — needs an outbox, and #331/#336 record that the existing outbox marks dropped events as published.

I would take the first, but it is a billing decision and should be made deliberately rather than inherited from whichever is easiest to code.

## Acceptance

- [ ] `UpdateSubscriptionParams` carries `TrialEnd`, and extending a trial cannot change the price
- [ ] The failure ordering above is decided explicitly and written down before implementation
- [ ] `409 stripe_managed` is replaced by a working path, not merely removed
- [ ] The value sent to Stripe is asserted as an **exact integer** in a test, not merely that a call occurred — a stub returns the zero value for a field nobody set
- [ ] Verified against a real Stripe test-mode subscription, since production has 0 subscription rows and cannot exercise this

mahesh-sangawar:
Closes #358.

`POST /admin/billing/trials/{store_id}/extend` refused every card-backed trial with `409 stripe_managed`. It now moves the trial end in Stripe and locally, and `GET /admin/billing/trials` can list those trials so the endpoint is reachable at all.

## The two decisions #358 asked to be made deliberately

**Stripe is called first and is the source of truth.** A Stripe failure rolls the transaction back and writes nothing locally. Only a failed COMMIT can diverge, and then Stripe is *ahead* — the merchant is billed later than the console shows, never earlier. The reverse ordering fails in the unsafe direction. The outbox option was rejected: #336 records that the publisher marks dropped events as published.

**The row lock is held across the Stripe call.** `SELECT … FOR UPDATE` on `store_subscriptions`, held across the network call under a 10s timeout. An external call inside a transaction is normally an anti-pattern; here it is what removes the window in which the subscription could convert while Stripe is in flight. Measured against real Stripe: **832ms – 1.15s** held, against the 10s bound.

## Verified against real Stripe, not only against tests

Production holds **0** `store_subscriptions` rows, so it cannot exercise any of this — a deploy proves only that the route is mounted and refuses unsigned callers. Against Stripe test mode, through our own client and confirmed independently via raw REST:

| claim | result |
|---|---|
| `trial_end` receives the exact integer sent | PASS |
| `billing_cycle_anchor` moves with it | PASS — previously an unverified SDK doc-comment claim |
| the subscription's price is unchanged | PASS — the re-pricing hazard, disproven for real |
| status stays `trialing` | PASS |

`internal/billing/trial/extend_stripelive_test.go` (build tag `stripelive`, invisible to every normal build and to CI) runs the full `Extender` against real Stripe **and** real Postgres in one transaction, and reports the lock's real held duration. Run it with `STRIPE_TEST_KEY` and `TEST_DATABASE_URL` set.

## What changed

- `stripe.UpdateSubscriptionParams` gains `TrialEnd`; `Items` is sent **only** when a `PriceID` is set, so extending can no longer re-price. `UpdateTrialEnd` is a narrow wrapper with no `PriceID` field, so the extend path cannot acquire one later.
- `trial.Extend` becomes a method on `trial.Extender` carrying the Stripe dependency. A nil updater reproduces today's refusal exactly, and `NewExtender` normalises a **typed** nil to a true nil — the shape that shipped as a Critical in #288.
- New refusals: `409 stripe_state_conflict`, `400 trial_end_too_far`, `400 trial_end_not_after_stripe`, `500 stripe_applied_local_write_failed`, `502 stripe_unavailable`. `409 stripe_managed` survives, narrowed to "no Stripe configured on this build".
- Response and audit row disclose `stripe_subscription_id`, the trial end **as Stripe reports it back**, and `billing_anchor_moved`.
- `GET /admin/billing/trials?include_stripe_managed=true` lists card-backed trials; every row carries `stripe_managed`. The default is unchanged, and `CountExpiring` — which backs #282's `trials_expiring` KPI — is untouched.

## Two findings the whole-branch review caught that no task-level review could

**Critical: the Stripe idempotency key could make a real extension a silent no-op.** The key was `(storeID, trial_end)` and the handler's scoped key never reached the domain. Extend to Dec 1 → correct to Nov 1 → back to Dec 1 within 24h replays the first call's key with identical parameters, so Stripe returns the cached response and performs nothing, while we write Dec 1 locally and report success. Root cause was a contradiction in the spec itself, which required both an identical method signature and a key derived from the handler's — mutually impossible, and invisible to any single task. The spec now records the contradiction rather than hiding it.

**Important: an "extend" could silently shorten a card-backed trial.** Nothing compared the requested date against Stripe's *current* `trial_end`. Now refused, naming both dates. **Consequence worth knowing: shortening a card-backed trial is no longer possible through this endpoint.** The card-less path is unchanged.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `go vet -tags=integration ./...`, `go vet -tags=stripelive ./internal/billing/trial/` — all clean
- [x] `internal/billing/trial` and `internal/handlers/platformadmin` green under `-tags=integration -p 1 -count=1`
- [x] `internal/subscription/planchange`'s pre-existing failure set measured 5 times against `origin/main` — the same 9 `store_subscriptions_store_id_fkey` fixture-drift failures (#317), never 10. That suite never reaches the plan-change logic, so the regression proof for making the price swap optional lives in `internal/billing/stripe`'s unit tests, which do run.
- [x] Both golden fixtures behaved correctly: `trial_extend_response.json` **unchanged** (the new fields are `omitempty`, so a card-less response carries none of them), `billing_trials_response.json` changed by exactly the added `stripe_managed` key.
- [x] Every fix proven by a mutation that fails for the reason it names.
- [ ] Not verifiable anywhere: production has 0 subscription rows. This needs Stripe test mode, which is what the `stripelive` test and `scripts/verify-358-stripe.sh` are for.

## Follow-ups, not in this PR

- **#366** — production's Stripe billing key is `sk_test_…`. Everything here becomes real money the moment it is swapped, with no code change to mark the transition.
- After an `ErrStripeAppliedLocalWriteFailed` divergence, Stripe already holds the target date, so a corrective retry at that date is refused and the local row stays behind; reconciliation is manual. Letting an equal date skip the Stripe call and heal locally would close it — deliberately not done in the last commit of this branch.
- `ErrStripeAppliedLocalWriteFailed`'s signal is best-effort: a process death between Stripe's reply and the return emits nothing.
- The reminder cron re-claims a cleared slot if it interleaves with an extension (pre-existing; higher stakes now the date is a real Stripe charge).


mahesh-sangawar:
## On the red `Secrets / Secret scan`

Investigated rather than waved through. **No credential is leaked**, and `main` will be clean after merge.

**What gitleaks flagged** (v8.30.1, rule `curl-auth-user`, 2 findings):

- `services/marketplace-api/scripts/verify-358-stripe.sh:81` — `curl -sS -u "$KEY:"`
- `docs/superpowers/plans/2026-08-26-trial-extend-stripe.md:1691` — the same line, quoted in the plan

Both take the key from an environment variable; nothing is hardcoded. The rule is new in 8.30.x, which is why a local 8.21.2 scan showed the branch clean — worth knowing for the next person who checks a scan result locally and believes it.

**Fixed at the cause, not baselined** (`88bf8600`). Passing a credential via `-u` puts it in the process's `argv`, where any user on the machine can read it with `ps`. That is a real weakness in a script whose whole purpose is careful key handling — it already refuses to write the key to a file or echo it. The helper now feeds curl its credential on stdin:

```sh
api() {
  local path="$1"; shift
  printf 'user = "%s:"\n' "$KEY" | curl -sS --config - "https://api.stripe.com/v1/$path" "$@"
}
```

An allowlist entry would have kept the weakness and silenced the signal.

**Why the check stays red anyway:** gitleaks scans git *history*, and the two commits that introduced those lines still exist on this branch. A later commit cannot unwrite them. Since this repo squash-merges, `main` receives a single commit containing the fixed tree, so the offending lines never reach `main` and its scan stays green. Verified after merge rather than assumed.

The alternative — rewriting this branch's history — would destroy the 18-commit record that makes this branch reviewable, to fix a condition that does not survive the merge.

mahesh-sangawar:
## Correction to my comment above

I claimed that squash-merging alone would leave `main` clean. **That was wrong, and CI proved it**: the first post-merge run on `main` (`6a0cd3b7`) failed secret scan too.

The reason: CI fetches **all** refs (`+refs/heads/*:refs/remotes/origin/*`), and I merged with `--delete-branch=false`, so `feat/358-trial-extend-stripe` still existed on the remote. gitleaks scans history reachable from any ref, so it still reached the two commits carrying the `curl -u` line. Squashing controls what lands on `main`'s own history; it does nothing about a branch ref that still exists beside it.

**Fixed by deleting the remote branch.** The next run on `main` (`7a17819c`) passes secret scan. The merged PR keeps its full 20-commit list on GitHub, so nothing was lost by deleting the ref.

Recording it because the original explanation was confident, specific, and wrong — the kind of claim that sends the next person looking in the wrong place.

PR #380, opened 2026-08-26
feat(platform-admin): publish the lifecycle reason codes (§8.8)

mahesh-sangawar:
Closes the product half of tesserix-home#345.

## The gap

§8.3 requires a reason code on suspend and unsuspend so an audit row says *why* and not only *what*. This service complies and always has — seven suspend codes, four deliberately different unsuspend ones, an unrecognised code refused with §4.4's `invalid_reason_code`.

**Nothing exposed the list.** The codes were a Go var in `tenant_lifecycle.go`, reachable only by opening this file, so the console shipped a hand-copied duplicate. A copied vocabulary drifts in two directions and only one is loud: offering a **retired** code is refused where someone sees it, missing an **added** one is silent — the option simply is not there and the operator picks the nearest wrong reason, which lands on an audit row and is never questioned again.

## What this adds

`GET /admin/lifecycle/reason-codes` → `{"data": {"suspend": [{"code","label"}], ...}}`, per the new contract §8.8 (tesserix-home).

**It serves all four verbs, not two.** §8.8 requires `suspend` and `unsuspend` of every implementer and permits more, and this service has four consequential verbs — `trial_extend` and `purge` have their own sets. Publishing only the two the console needs today would have earned a second migration the first time anyone built a trial-extension form.

## What it deliberately does not change

The `[]string` code lists remain the single authority on membership and order. The new label map is **consulted, never enumerated**, so `isKnownReasonCode` and the `allowed` array in the §4.4 error bodies are byte-identical to before. Publishing a vocabulary must not become an opportunity to change what the vocabulary is.

**Labels are nested per verb, and that nesting is load-bearing.** `operator_error` means "suspended in error" on unsuspend, "correcting a mistaken earlier extension" on trial_extend, and "a tenant created in error" on purge; `fraud` and `legal` likewise differ between pausing a tenant and destroying one. A flat code→label map would have shown an operator the wrong sentence next to the right code — worse than showing the bare code.

**The route mounts unconditionally**, unlike the writes it describes. Those refuse to mount without an audit path because an unattributed write is worse than an absent one. This is a read of a compile-time constant: it depends on nothing, and gating it on the writes being wired would answer 404 in exactly the deployment where an operator most needs to know what the vocabulary is.

## Follow-up, deliberately not here

`admin-conformance.json` (mark8ly#290) is **not** in this PR. Declaring the two new endpoint ids requires `@tesserix/admin-conformance@0.4.0`, which is still in review as tesserix/design-system#33 — `parseEndpoints` throws on an id the installed package does not know, so the file cannot land first.

## Test plan

- [x] `go build ./...`, `go vet` clean
- [x] Full `platformadmin` package suite passes
- [x] Endpoint shape asserted by decoding **raw JSON** into an anonymous struct, not the producer's type — a wrong or missing json tag would otherwise be invisible
- [x] Drift guard fails on a code with no label. **Verified by adding one**: `verb "suspend" code "drift_probe" fell back to its own code as a label`
- [x] A test asserts the shared `operator_error` reads differently on all three verbs that use it
- [x] A test asserts suspend and unsuspend stay disjoint

### DIFF
--- a/services/marketplace-api/internal/handlers/platformadmin/lifecycle_reason_codes.go
+++ b/services/marketplace-api/internal/handlers/platformadmin/lifecycle_reason_codes.go
@@ -1,0 +1,153 @@
+// services/marketplace-api/internal/handlers/platformadmin/lifecycle_reason_codes.go
+package platformadmin
+
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+)
+
+// ReasonCode pairs a wire code with the words an operator reads.
+//
+// The code is what crosses the wire, lands in an audit row and is matched
+// exactly by the validators in this package. The label is presentation and
+// may be reworded freely; renaming a code may not.
+type ReasonCode struct {
+	Code  string `json:"code"`
+	Label string `json:"label"`
+}
+
+// LifecycleReasonCodesResponse is contract §8.8's body.
+//
+// A map rather than four named fields: §8.8 requires `suspend` and
+// `unsuspend` of every implementer and permits more, and a product's set of
+// consequential verbs is its own. Adding one here should not need a struct
+// field, and a console reading it should not need a new build to see one.
+type LifecycleReasonCodesResponse struct {
+	Data map[string][]ReasonCode `json:"data"`
+}
+
+// reasonCodeLabels supplies the human wording for each code, per verb.
+//
+// **Per verb, not global, and that is the whole reason this is a nested map.**
+// `operator_error` means "suspended in error" on unsuspend, "correcting a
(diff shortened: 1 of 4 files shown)


mahesh-sangawar:
Pushed a second commit: `gofmt` on `internal/outbox/platform_list_integration_test.go`.

It is **not** part of this change — the file is unformatted on `origin/main` already, so the `Go (marketplace-api) / Go quality` job is red for every PR opened off main right now, not just this one. Two lines of comment alignment. Verified pre-existing by running `gofmt -l` against a clean `origin/main` checkout.

Happy to split it into its own PR if you would rather keep this one single-purpose.

PR #417, opened 2026-08-28
fix(platformadmin): declare inbox and serve the contract envelope it requires (#415)

### LINKED ISSUE #415: admin-conformance.json does not declare inbox, which is mounted and working
`admin-conformance.json` declares eight endpoints:

```json
"audit-logs", "health", "kpis", "entities": ["tenants","users"],
"billing/subscriptions", "billing/trials", "tenant-lifecycle",
"lifecycle/reason-codes"
```

**`inbox` is absent**, and it is implemented: `platformadmin/routes.go` mounts `NewInboxHandler` and `NewInboxActionsHandler`, with `inbox.go`, `inbox_actions.go`, `inbox_action_migration.go` and `inbox_idempotency.go` behind them. #281 closed on exactly that work.

## Why the declaration matters more than it looks

The contract is **declared, not discovered** — §D4 of the console integration design: *"The rail renders only declared contracts. Mark8ly's rail is its declaration."* So a working endpoint that is not declared is, to a consumer reading the declaration, not there. mark8ly's queue can be simultaneously correct in code and invisible by contract.

That is a worse failure than a broken endpoint, because nothing errors. The console's estate Inbox renders declared sources; an undeclared one produces an empty queue and no diagnostic.

## Check the neighbours while in here

Two others are mounted and also undeclared — `outbox`, `email-sends`, `notifications`, `tickets`, `break-glass`, `entities/conversions`, `onboarding-funnel`. Some of those are deliberately not contract endpoints (they are mark8ly-specific reads the console consumes directly), and some may be omissions like `inbox`. **The distinction should be recorded, not inferred.** A reader today cannot tell which absences are decisions.

Note `inbox` also takes a shape argument rather than a bare `true` — Kora declares `"inbox": { "slaDeclared": false }`. mark8ly has real SLAs on at least one queue (`sea_manual_review` carries a five-business-day `sla_due_at` that pauses a subscription clock), so its declaration is not the same value.

## Also: implemented ≠ answering

Every route in `routes.go` mounts behind `if deps.X != nil`. So there are three distinct states — implemented, declared, and *wired in the deployed environment* — and this file only speaks to the second. Worth a line in the file or its doc saying so, because a green conformance run against an unwired dependency is exactly the "passing while unusable" trap `docs/admin-contract.md` names for Kora's search endpoint.

### Acceptance

- [ ] `inbox` declared, with the SLA shape its queues actually warrant
- [ ] Every mounted platform route is either declared or has a recorded reason it is not
- [ ] A test fails when a route is mounted without a corresponding declaration decision

mahesh-sangawar:
Closes #415.

mark8ly's estate queue was implemented, deployed, and invisible by contract. Declaring it turned out to require fixing it first.

## What #415 asked for, and what was actually true

The issue is right that `inbox` is missing from `admin-conformance.json`. Two of its three acceptance items change shape once you read the conformance suite's source, and the third finding is the one that mattered.

**The declarable vocabulary is closed.** `design-system/packages/admin-conformance/src/contract.ts:13-126` defines exactly nine ids. mark8ly declared eight. So `inbox` was the only real gap — and the issue's "check the neighbours" list (`outbox`, `email-sends`, `notifications`, `tickets`, `break-glass`, `conversions`, `onboarding-funnel`) **cannot be declared at all**: `declaration.ts` throws on an unrecognised key and fails the entire run. Their absence is one structural fact, not seven unrecorded decisions. That is what `docs/admin-conformance.md` now records.

**`slaDeclared` must be `false`, not the different-from-Kora value the issue expected.** `checks.ts:206-220` requires `due_at` on *every* item when an SLA is declared. Only one of mark8ly's five inbox kinds sets it (`sea_review.go:79`); the other four never do, and `erasure.go:13` documents that as a principled refusal — its table has no due column, and deriving a statutory GDPR deadline in a read endpoint would be inventing policy in the wrong place. `true` fails; `false` understates. Raised upstream as design-system#36.

**`GET /admin/inbox` did not serve the contract's envelope.** This is the finding that changed the work. `contract.ts:34` fixes inbox at `envelope: "items-total"`, and `envelope.ts:64-89` requires **exactly** the keys `items` and `total` — extra top-level keys fail too. mark8ly emitted the house `{data, pagination, degraded?}`. So declaring `inbox` without fixing the body would have swapped one nightly failure for another.

## What changed

- **`GET /admin/inbox` now emits `{ items, total }`.** `items` is `[]`, never `null`, when empty.
- **`degraded` moved to an `X-Inbox-Degraded` response header**, comma-separated, set only when non-empty. It names the kinds that could not be reached, and dropping it silently would let a partially-failed aggregation read as a healthy short queue — a worse failure than an error. The contract governs the body, so a header keeps the signal without breaking "exactly { items, total }". The response struct documents where it went, so the next reader finds out from the code.
- **`inbox` declared** as `{ "slaDeclared": false }`, with the file reordered into the contract's own id order.
- **`docs/admin-conformance.md`** carries the reasoning the JSON cannot hold — JSON has no comments — with file:line citations into both repos so every claim is checkable.
- **A guard test** reconciles the declaration against the real gin router in both directions: a contract-id route mounted but undeclared fails, and a declared id with no mounted route fails. `entities` is checked per subtype, because `tenants` and `users` mount on independent `Deps` fields. Unknown keys fail here rather than overnight against production.

## Sequencing — this is not the whole fix

The nightly job reads a **different copy** of the declaration: a ConfigMap rendered from `charts/apps/mark8ly-marketplace-api-admin/values.yaml` in `tesserix-k8s`, not this repo's file. The chart documents the duplication and asks for manual sync (mark8ly#290).

So this PR alone does not make the job green. The order is:

1. **This PR merges and deploys**, so production serves `{ items, total }`.
2. **Then** the chart change lands (`tesserix-k8s`, branch `feat/mark8ly-declare-inbox-conformance`), declaring `inbox` in the copy the job reads.

Reversing that order fails the run on the envelope. Until step 2, the job fails on "answers 200 but is not declared" — which is its current state anyway, since `/admin/inbox` started answering 200 in production eight hours after the last run.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `go vet -tags=integration ./...` — all exit 0
- [x] `go test ./... -count=1` from the service root — exit 0
- [x] Response body asserted to have the **exact** key set `{items, total}`, so a future added field fails here rather than overnight
- [x] Empty page returns `200` with `"items": []`, not `null`
- [x] `X-Inbox-Degraded` present with the right comma-separated value when degraded, and **absent** — not empty-string — when not
- [x] Guard proven to fail when `inbox` is removed from the declaration, and when a declared `entities` subtype is unmounted
- [x] Existing behaviour preserved: filters, paging, 400-on-unknown-kind, 400-on-deep-page, 500-on-all-sources-failed
- [x] Rendered chart ConfigMap validated: nine known ids, no unknown keys
- [x] Cross-repo drift guard: fails when the repo and chart declarations disagree; skips (naming the path) when the sibling checkout is absent, as it is in CI
- [ ] A conformance run confirms green — triggered manually after step 2, not waiting for 06:20 UTC

## Notes

The last run (2026-08-27 06:20Z) failed on three `entities/tenants` §8.9 "has no label" checks. That was fixed by `c701f82f` and is already deployed; unrelated to this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018jFmanqTbbCYUgqZRGgQT7

### DIFF
--- a/admin-conformance.json
+++ b/admin-conformance.json
@@ -2,15 +2,18 @@
   "slug": "mark8ly",
   "contractVersion": 2,
   "endpoints": {
-    "audit-logs": true,
-    "health": true,
     "kpis": true,
+    "inbox": {
+      "slaDeclared": false
+    },
+    "audit-logs": true,
     "entities": {
       "types": [
         "tenants",
         "users"
       ]
     },
+    "health": true,
     "billing/subscriptions": true,
     "billing/trials": true,
     "tenant-lifecycle": true,
--- a/admin-conformance.json
+++ b/admin-conformance.json
@@ -4,7 +4,9 @@
   "endpoints": {
     "kpis": true,
     "inbox": {
-      "slaDeclared": false
+      "slaKinds": [
+        "sea_manual_review"
+      ]
     },
     "audit-logs": true,
     "entities": {
--- a/docs/admin-conformance.md
+++ b/docs/admin-conformance.md
@@ -80,7 +80,8 @@ inbox kinds carries a real, meaningful SLA. `sea_manual_review` reads
 (`services/marketplace-api/internal/inbox/sea_review.go:12-15`) states that
 any row entering the queue immediately pauses the 14-day validation clock on
 the associated subscription, under a 5-business-day SLA
-(`sla_due_at`, populated at `sea_review.go:72`, `DueAt: &due`).
+(`sla_due_at`, read into `due` at `sea_review.go:72` and assigned to
+`DueAt: &due` at `sea_review.go:79`).
 
 But `slaDeclared: true` is a per-queue, not a per-kind, promise:
 `design-system/packages/admin-conformance/src/checks.ts:206-220` requires
--- a/docs/admin-conformance.md
+++ b/docs/admin-conformance.md
@@ -1,0 +1,135 @@
+# Admin conformance declaration
+
+This file explains the reasoning behind `admin-conformance.json` — the
+declaration `mark8ly`'s nightly conformance run (design-system's
+`admin-conformance` suite) checks against production. The JSON file itself
+is parsed with `JSON.parse` and cannot carry comments, so every "why" that
+isn't self-evident from the key/value pair goes here instead. Issue: #415.
+
+## The declarable vocabulary is closed to nine ids
+
+`admin-conformance.json`'s `endpoints` map may only contain keys drawn from
+the Product Admin Integration Contract's fixed id list. That list is defined
+once, in the design-system monorepo, and is exactly nine entries:
+
+`design-system/packages/admin-conformance/src/contract.ts:13-126` —
+`ENDPOINTS` (and the derived `ENDPOINT_IDS`) define exactly:
+
+1. `kpis`
+2. `inbox`
+3. `audit-logs`
+4. `entities`
+5. `health`
+6. `billing/subscriptions`
+7. `billing/trials`
+8. `tenant-lifecycle`
+9. `lifecycle/reason-codes`
+
+There is no tenth id, and there is no way to add a private one: the parser
+that reads this file, `design-system/packages/admin-conformance/src/declaration.ts`
+(`parseEndpoints`), **throws** on any key that is not one of the nine —
+`unknown endpoint ${key}; valid ids are: ...` — which fails the *entire*
+conformance run, not just the offending entry. A product cannot partially
+declare; an unrecognised key takes every other declared endpoint down with
+it. This is why the file must stay strictly inside the nine ids and nothing
+else, and why the guard in
+`services/marketplace-api/internal/handlers/platformadmin/conformance_declaration_test.go`
+independently mirrors this same closed set and asserts every key in this
+file is a member of it — a typo here should fail in mark8ly's own CI, not
+overnight against production.
+
+## The seven mounted-but-undeclarable routes
+
+mark8ly's platform-admin surface (`services/marketplace-api/internal/handlers/platformadmin/routes.go`)
+mounts several `/admin/*` reads that the console consumes directly but that
+have **no corresponding id in the nine above**. Because the vocabulary is
+closed and an unknown key throws (see above), these routes structurally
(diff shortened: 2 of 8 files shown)


mahesh-sangawar:
Two notes while this is still open.

**1. The kind names in `docs/admin-conformance.md` are shorthand, not the wire values.**

The doc lists the five kinds as `sea_manual_review`, `arbitrage`, `migration_fastpath`, `onboarding`, `erasure`. Four of those are not what goes on the wire — `internal/inbox/provider.go` defines `arbitrage_appeal`, `migration_fast_path`, `onboarding_stalled` and `erasure_request` (only `sea_manual_review` matches).

Harmless as prose today. It stops being harmless the moment anyone copies a name from this paragraph into a declaration, because the checker matches on `item.kind` — a near-miss would silently check nothing rather than fail, which is the failure mode this whole file exists to prevent.

**2. This PR's `slaDeclared: false` decision has been superseded upstream.**

design-system#36 is fixed and merged: `@tesserix/admin-conformance` gains `slaKinds`, so a mixed queue can be honest per kind rather than choosing between a boolean that fails and one that understates.

```jsonc
"inbox": { "slaKinds": ["sea_manual_review"] }
```

`due_at` is then required only of items whose `kind` is listed; the other four are free to omit it, including `erasure_request` — whose omission `erasure.go:13` documents as a deliberate refusal, and which stays a refusal under this shape. A declared kind absent from the sampled page reports as a `skip`, not a `pass`.

**Do not apply this yet.** npm still serves `0.6.0`, where `slaKinds` is an unknown option and `parseDeclaration` *throws* — failing the entire run rather than one entry. design-system's release PR has not merged. Once `0.7.0` publishes, the change is one line here plus replacing this PR's "Why `slaDeclared` is `false`" section, and it belongs on this branch rather than a follow-up — the section currently argues at length for a value that is about to stop being the right one.

Issue #484, opened 2026-08-30
Contract half of #177: drop the 8 legacy warehouse_* columns from shipping_carrier_configs


mahesh-sangawar:
## Now unblocked

Migration 000095 was deliberately the **expand** half of an expand/contract
migration. Its own comment says so:

> Nothing is dropped here on purpose: the running code still reads the old
> columns, so this migration must be a no-op for it. The contract half (drop
> the columns) only lands once every reader goes through warehouse_id.

#480 made every reader go through `warehouse_id`. The precondition is met.

## What is left on `shipping_carrier_configs`

Eight columns, still written and still read as a fallback:

```
warehouse_name, warehouse_line1, warehouse_line2, warehouse_city,
warehouse_region, warehouse_postal, warehouse_country, warehouse_phone
```

## Why this is not just a DROP

The fallback path in #480 is load-bearing, not defensive. Three readers use the
legacy columns whenever `warehouse_id IS NULL`:

- `internal/handlers/admin/shipments.go` — `resolvePickupAddress`
- `internal/handlers/storefront/shipping_rates.go` — `resolveWarehouseAddress`
- `internal/handlers/admin/settings.go` — `resolveWarehouseForSync`

So before dropping anything:

- [ ] Confirm **zero** rows have `warehouse_id IS NULL AND warehouse_name <> ''`
      in production. 000095's backfill skipped rows with a blank name (a blank
      name was never a usable warehouse), and #480's write path only sets
      `warehouse_id` on save — so a config not re-saved since #480 shipped
      still relies on the fallback
- [ ] Backfill any stragglers found by the check above
- [ ] Remove the fallback branch from all three resolvers, and the
      `Warehouse*` fields from `shipping.CarrierConfig`,
      `ShippingCarrierConfigRow` and `carrierConfigRow`
- [ ] Stop writing the columns in `admin/settings.go`
- [ ] Then the migration to drop them

## One loose end

`shipping.UpsertCarrierConfig` (`internal/shipping/repository.go:312`) writes
the legacy columns and does **not** maintain `warehouse_id`. It has no
production caller today — only the interface, the implementation, and a test
fake — so it causes no divergence now. If anyone wires it up before this
contract half lands, it must be taught about `warehouse_id` first, or it will
silently write an address the read path ignores.

Follow-up to #177 / #480.

PR #485, opened 2026-08-30
feat(shipping): collect warehouse contact person and email (#483)

### LINKED ISSUE #483: Collect warehouse contact_person and email — Delhivery registration still fails without them
## The gap

`shipping.Warehouse` has carried `ContactPerson` and `RegisteredName` since the
interface was written (`internal/shipping/carrier.go:48`), and until #480 **no
caller had ever set them**. That is the root of the Delhivery label failure that
opened #177:

```
ClientWarehouse matching query does not exist
```

Delhivery's `/clientwarehouse/create/` takes `contact_person` + `email`. The old
`warehouse_*` columns on `shipping_carrier_configs` had no column for either, so
`shipments.go` fell back to the **buyer's order email** for `pickup_email` —
which was wrong, and was already flagged in a TODO there.

## What #480 fixed, and what it did not

Migration 000095 gave the `warehouses` table `contact_person` and `email`
columns, and #480 made both flow through to the carrier at the two sites that
build a `shipping.Warehouse`:

- `internal/handlers/admin/shipments.go` — the `WarehouseNotRegisteredError`
  auto-registration retry
- `internal/handlers/admin/settings.go` — `resolveWarehouseForSync`

**But nothing collects them from the merchant.** The admin shipping-settings
request struct (`shippingUpsertRequest`) has no `contact_person` / `email`
field, and the UI has no input for them. So in practice both columns are still
empty for every store, and registration still fails for merchants who need it.

This is worth its own issue precisely because #480 makes it *look* addressed:
the plumbing is there and tested, the data is not.

## What is needed

- [ ] Add `contact_person` and `email` to `shippingUpsertRequest` and the
      settings response, and pass them into the `warehouse.Upsert` call added
      in #480
- [ ] Add the two fields to the admin shipping settings form (~11 UI files
      touch warehouse today)
- [ ] Decide the fallback when a merchant leaves them blank. Delhivery defaults
      both to `Name`, per the comment at `carrier.go:48` — confirm that is
      actually acceptable to Delhivery rather than assumed
- [ ] Drop the buyer-email fallback for `pickup_email` in `shipments.go` once a
      real warehouse email exists (it is currently still `o.CustomerEmail`)
- [ ] Verify against Delhivery test-mode credentials that registration now
      succeeds — the whole point is that this failed in production on
      2026-07-17

## Note on the migration

`email` and `contact_person` are **nullable** in 000095 and left NULL by its
backfill, so every pre-existing row has them empty. Any UI that treats them as
required has to cope with that on first edit.

Follow-up to #177 / #480.

mahesh-sangawar:
Merchants can now enter a pickup **contact person** and **contact email**, and
those values reach the carrier. Closes #483.

## What changed

**Backend** — `contact_person` / `email` added to the shipping settings request
and response. They are stored ONLY on the `warehouses` row (there are no legacy
`shipping_carrier_configs` columns for them), so the response sources them by
resolving `warehouse_id`. Both are optional: every existing row has them NULL,
and requiring them would break saves for every current store.

**Frontend** — two optional inputs in the existing Warehouse address section of
`ShippingConfigForm.tsx`, plus the two fields on the request/response types.

**Label path** — `shipments.go` now prefers the warehouse's own email for the
pickup contact, falling back to the buyer's email only when the warehouse has
none.

## Correcting the premise of #483

I wrote that issue claiming the unset `ContactPerson` was "the root of" the
`ClientWarehouse matching query does not exist` failure. **That was wrong**, and
reading `delhivery.go` while implementing this is what showed it:

```go
// delhivery.go:769
contactPerson  := firstNonEmpty(wh.ContactPerson, wh.Name)
registeredName := firstNonEmpty(wh.RegisteredName, wh.Name)
```

`contact_person` has always been populated — it defaults to the warehouse name,
and the comment above it says exactly that. So it was never missing, and this PR
does **not** fix that registration error. Whatever caused it lies elsewhere
(most likely the warehouse simply never having been created carrier-side, which
the existing `WarehouseNotRegisteredError` auto-registration retry addresses).

**What this PR does genuinely fix** is the email. Delhivery is sent `email` only
when non-empty, and the value being sent was `o.CustomerEmail` — the **buyer's**
address — as the warehouse's pickup email. That is both wrong and a small
privacy leak of customer data to the carrier. A merchant can now set a real one.

I will post the same correction on #483 so the issue does not keep asserting a
cause that is not real.

## Also here: a test-infrastructure fix

`pkg/testdb.openOrSkip` opened a new pool per test and **never closed it**, with
no bound on max-open connections. Packages accumulated pools until the shared
server refused them:

```
FATAL: sorry, too many clients already (SQLSTATE 53300)
```

Adding four tests to `internal/handlers/admin` was enough to cross the ceiling,
and the failure surfaces on whichever test happens to run at that moment — so it
reads as that test's own bug. It is not. Pools are now bounded and closed on
cleanup (registered before `NewDB`'s truncate cleanup, which is LIFO, so the
connection is still alive for the truncate that needs it).

This is separate from `cmd/testdb-leakcheck`, which catches leftover *rows*.

## Testing

```
TEST_DATABASE_URL='postgres://dev:dev@<host>:5432/marketplace_db?sslmode=disable' \
  go test -tags=integration -p 1 -count=1 ./internal/handlers/admin/... ./internal/warehouse/... ./internal/handlers/storefront/...
```

All green. New cases: values persist on the warehouses row; they round-trip on
the response; omitting them does not wipe stored values; a row with NULL contact
columns still saves; the label path prefers the warehouse email and falls back
to the buyer's.

`apps/admin` typecheck introduces no new errors — the two `@tesserix/otto-widget`
failures are present on `main` and unrelated.

## Not in scope

Dropping the legacy `warehouse_*` columns is #484.

### DIFF
--- a/apps/admin/components/settings/ShippingConfigForm.tsx
+++ b/apps/admin/components/settings/ShippingConfigForm.tsx
@@ -253,7 +253,7 @@ export function ShippingConfigForm({
             />
             <p className="text-xs text-[color:var(--ink-900)]/40 mt-1">
               Used for shipping label pickup notifications. Falls back to the
-              customer's email if left blank.
+              customer&apos;s email if left blank.
             </p>
           </Field>
         </div>
--- a/apps/admin/components/settings/ShippingConfigForm.tsx
+++ b/apps/admin/components/settings/ShippingConfigForm.tsx
@@ -76,6 +76,13 @@ export function ShippingConfigForm({
     phone: existing?.warehouse_phone ?? "",
   });
 
+  const [warehouseContactPerson, setWarehouseContactPerson] = useState(
+    existing?.warehouse_contact_person ?? "",
+  );
+  const [warehouseEmail, setWarehouseEmail] = useState(
+    existing?.warehouse_email ?? "",
+  );
+
   const [error, setError] = useState<string | null>(null);
   const [success, setSuccess] = useState(false);
   const [pending, startTransition] = useTransition();
@@ -101,6 +108,8 @@ export function ShippingConfigForm({
         warehouse_postal: address.postal || undefined,
         warehouse_country: address.country || undefined,
         warehouse_phone: address.phone || undefined,
+        warehouse_contact_person: warehouseContactPerson || undefined,
+        warehouse_email: warehouseEmail || undefined,
         auto_schedule_pickup: autoSchedulePickup,
         default_pickup_slot_start: defaultPickupSlotStart,
         // Slot end mirrors the start + 4h — kept implicit because the UI
@@ -218,6 +227,36 @@ export function ShippingConfigForm({
           disabled={pending}
           idPrefix={`${provider}-wh`}
         />
+        <div className="grid gap-4 sm:grid-cols-2">
+          <Field label="Contact person (optional)" htmlFor={`${provider}-wh-contact`}>
+            <input
+              id={`${provider}-wh-contact`}
+              type="text"
+              value={warehouseContactPerson}
+              onChange={(e) => { setWarehouseContactPerson(e.target.value); setSuccess(false); }}
+              disabled={pending}
+              className={inputClass}
+            />
+            <p className="text-xs text-[color:var(--ink-900)]/40 mt-1">
+              Who the courier should ask for when picking up. Falls back to the
+              warehouse name if left blank.
+            </p>
+          </Field>
+          <Field label="Contact email (optional)" htmlFor={`${provider}-wh-email`}>
+            <input
+              id={`${provider}-wh-email`}
+              type="email"
+              value={warehouseEmail}
+              onChange={(e) => { setWarehouseEmail(e.target.value); setSuccess(false); }}
+              disabled={pending}
+              className={inputClass}
+            />
+            <p className="text-xs text-[color:var(--ink-900)]/40 mt-1">
+              Used for shipping label pickup notifications. Falls back to the
+              customer's email if left blank.
+            </p>
+          </Field>
+        </div>
       </fieldset>
 
       {/* Pickup automation — Delhivery only, but the form is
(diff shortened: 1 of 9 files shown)


mahesh-sangawar:
## Added after the PR was opened: a real bug in the blank-preserve logic

Blank `contact_person`/`email` means **unchanged**, not **clear** (unlike the other `warehouse_*` fields, which stay full-overwrite). `Upsert`'s `ON CONFLICT DO UPDATE` overwrites both columns unconditionally, so the handler has to resolve what to write *before* calling it.

The first version resolved that fallback by `existing.WarehouseID` — which goes **stale**, and interacts badly with behaviour introduced in #480:

1. merchant saves warehouse "Main" with a contact person
2. merchant clears `warehouse_name` → #480 nils `warehouse_id` on the config, while the `warehouses` row itself survives (other carriers may still use it)
3. merchant re-saves the same name "Main" → lookup by the now-nil `warehouse_id` finds nothing, falls back to blank, and **silently wipes the stored contact fields**

Every step is reachable from the admin UI.

Fixed by resolving the prior row by `(store_id, name)` — the same key `Upsert` itself conflicts on — via a newly exported `warehouse.Repository.ByStoreAndName`. `ErrNotFound` (a genuinely new name) is distinguished from real errors, which are logged.

`TestShippingUpsert_ClearingThenRestoringTheWarehouseNamePreservesContactFields` pins the sequence. I verified it **fails** against the old `warehouse_id` lookup and passes with the fix, so it is not vacuous.

## Also corrected

The `carrier.go` comment claimed Delhivery defaults *both* `ContactPerson` and `Email` to `Name`. Only `ContactPerson` does (`firstNonEmpty(wh.ContactPerson, wh.Name)`); `Email` is simply omitted when blank, because Delhivery validates the format and a name-as-email would fail it. Comment-only change, no behaviour difference.

## Known, not addressed here

- `toShippingResponse` does a per-config warehouse query — not an N+1 at today's carrier counts, but it would become one.
- `settings.go` is now ~1175 lines, past the repo's 800-line guideline. Mostly pre-existing; ~100 lines are from this change.

Issue #177, opened 2026-07-17
Multi-warehouse support (per-location inventory + order allocation)


mahesh-sangawar:
## Summary

Support more than one warehouse / pickup location per store, so orders can be
fulfilled from the location that actually holds the stock.

Today a store has exactly one warehouse, and it isn't even a first-class
entity — it's 8 `warehouse_*` columns duplicated onto every row of
`shipping_carrier_configs`. Configure Delhivery *and* CouriersPlease and the
merchant types the same address twice.

## Why this is smaller than it looks

The data model was already designed for it, then deliberately pinned to one
location for slice 1:

```go
// DefaultLocationID is the slice-1 single-warehouse location used for
// variant_stock seeding. Multi-warehouse lands in a later slice.
const DefaultLocationID = "00000000-0000-0000-0000-000000000001"
```
— `internal/product/service.go:44`

- `variant_stock` already has composite PK **`(variant_id, location_id)`** — the
  stock table is multi-location at the DB level today. All rows currently use
  the sentinel `00000000-0000-0000-0000-000000000001`.
- `UpdateVariantStockInTx` is *"the ONLY place in the whole service where we
  mutate `variant_stock.quantity`"* (guarded by a test) — one chokepoint to
  teach about locations.
- Only **7 call sites** pin `DefaultLocationID`, all under `internal/product/`.
- Storefront reads `product_variants.inventory_quantity`, which the
  `sync_variant_inventory` trigger rolls up from `variant_stock`. **Browse / PDP
  / cart need no changes** — total stock across warehouses stays the right
  number to show.

## Scope

- [ ] `warehouses` table (store-level), migrate the 8 columns off
      `shipping_carrier_configs` → `warehouse_id` FK
      **(split out — see the "cheap half" below, doable independently)**
- [ ] Backfill `variant_stock.location_id` sentinel → real warehouse id
- [ ] Admin: warehouse CRUD + per-location stock editing (~11 UI files touch
      warehouse today)
- [ ] `shipments.warehouse_id` — a shipment currently cannot record where it
      shipped from
- [ ] Rate calculation picks an origin warehouse
- [ ] Allocation rule (see below)
- [ ] Register each warehouse with Delhivery (`UpsertWarehouse` already does
      adopt-or-create via edit→create; blocked on #<contact_person /
      registered_name gap>)

## The actual hard part: allocation

When an order arrives, which warehouse ships it? This is a **product decision,
not an engineering one**, and it's where the time goes:

[...]

PR #515, opened 2026-09-01
docs(177): plan PR 5 — warehouse CRUD, archiving, per-location stock

mahesh-sangawar:
Plan only — no code. Refs #177.

PR 5 is the last piece before PR 6's sentinel backfill, and it carries the
proper fix for the warehouse-name trap: the carrier form's free-text address
becomes a picker over real warehouses.

## The part worth reading first

**The spec is now stale in a way that affects PR 6.** It says:

> Nobody is using warehouses yet. Production has zero warehouses and zero
> carrier configs.

As of today `the-bondi-store` has a warehouse named exactly **`Main Warehouse`**
(`52b67b6b-…`) and an active ShipEngine config bound to it. The spec has PR 6's
backfill *create* a warehouse per stock-holding store with that same name, and
`warehouses` is keyed `(store_id, name)` — so a naive INSERT either violates the
constraint, or an `ON CONFLICT DO NOTHING` silently leaves stock on the sentinel
while the code believes it migrated. The backfill must be idempotent against an
existing `Main Warehouse`, and tested against a database that already has one.

That is exactly the "what does the second call see?" failure the #177 review
rounds kept producing, so it is recorded at the top of the plan rather than
left to be rediscovered.

## Structure

Five slices, each its own PR: schema+repository, admin API, settings page,
carrier-form picker, per-location stock.

Two design points I want challenged if they are wrong:

1. **Delete vs Archive.** Per the spec a warehouse with *any* allocation
   history is archived, never deleted, with `ON DELETE RESTRICT` as the
   backstop — so `Delete` applies only to a warehouse with no history. The plan
   says the UI must expose one verb, not make the merchant choose.
2. **The partial unique index.** `(store_id, name)` has to become
   `WHERE archived_at IS NULL`, or archiving burns a name forever. This is the
   quiet failure in the slice and is called out as such.

## Cleanup the plan commits to

5d must **delete** `warehouse-validation.ts` (#508) and update
`shipping-readiness.ts` (#511) in the same slice — once the address moves to the
warehouse form, those read fields the carrier config no longer holds. Leaving
both is two validators for one rule, which diverge.

## Explicit non-goals

Not retiring `DefaultLocationID` (PR 6), not touching the allocator (PR 2), not
changing the `shipment` → list response shape.

### DIFF
--- a/docs/superpowers/plans/2026-09-01-multi-warehouse-05-warehouse-crud.md
+++ b/docs/superpowers/plans/2026-09-01-multi-warehouse-05-warehouse-crud.md
@@ -1,0 +1,178 @@
+# #177 PR 5 — Warehouse CRUD, archiving, and per-location stock
+
+Spec: `docs/superpowers/specs/2026-08-31-multi-warehouse-allocation-design.md`
+Depends on: PRs 1–4b (merged: #488, #490, #491, #492, #495)
+Blocks: PR 6 (sentinel backfill, retiring `DefaultLocationID`)
+
+---
+
+## Correction to the spec before anything else
+
+The spec's Risks section says:
+
+> Nobody is using warehouses yet. Production has zero warehouses and zero
+> carrier configs.
+
+**That is no longer true, as of 2026-09-01.** `the-bondi-store` now has:
+
+- a warehouse `52b67b6b-b85d-496d-9466-ee95fbf9f1ee`, named exactly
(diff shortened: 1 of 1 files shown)


mahesh-sangawar:
All of #177 5a–5e shipped, along with the sentinel backfill and the retirement of DefaultLocationID and the read-time tolerance. Verified in production: all 16,354 units across 380 rows moved off the sentinel onto the-bondi-store's pre-existing Main Warehouse, schema at 123. This plan PR has nothing left to carry.

PR #529, opened 2026-09-01
feat(warehouse): admin CRUD API with archiving, reorder and default (#177 PR 5b)

mahesh-sangawar:
Refs #177. **Stacked on #528** — merge that first; this PR's base is its branch, so the diff shown here is 5b alone.

Until now a warehouse was a *side effect*: saving a shipping carrier config upserted one behind the form, keyed on `(store_id, name)`. A merchant who typed a slightly different name got a second, stockless warehouse instead of an edit, allocation then reported nothing available, and the order never shipped. #505 and #508 made that visible. This makes the warehouse a thing the merchant manages directly, which is the actual fix.

## The API

`/api/v1/admin/stores/:storeId/warehouses` — `GET` (list, `?include_archived=true`), `POST`, `PATCH /:id`, `DELETE /:id`, `PUT /reorder`, `PUT /:id/default`. Read is `RoleAdmin`; every write is `RoleOwner`.

### Repository: id-keyed, because Upsert cannot serve this

`Upsert` is keyed on `(store_id, name)` and a rename through it lands on a *different row*, leaving the stock behind — the exact trap #177 exists to remove. So 5b adds `Create`, `UpdateByID`, `SetDefault` and `UnitsHeld`, all keyed on the id and scoped by `store_id` (an id alone is a guessable handle to another tenant's row). A live name collision is `409 warehouse_name_taken`; an **archived** name is free to reuse, which is what 000122's partial unique index buys.

### One verb, not two

The spec splits removal in half — archive when there is allocation history, delete only when there is none — and that distinction is real but is not the merchant's to make. `DELETE` decides: it deletes when it can, archives when it must, and reports which happened plus `units_remaining`. Every refusal `Delete` can raise is a reason to archive, not a reason to say no; refusing outright would leave a stocked warehouse permanently unremovable.

`units_remaining` is not decoration. Archiving does **not** move stock, and (per #528) those units stop being sellable the moment the allocator skips the row. 5c needs the number to warn before, or explain after.

### Reorder takes the complete set

A delta applied to a list that changed underneath — a warehouse archived in another tab — reorders the wrong rows and the caller never finds out. The handler compares against the live set and returns 400 instead.

## A guard I did not know existed

`TestAdminRouteParity` caught all six new routes as missing from the mobile admin table. Declared web-only, alongside `/settings`: warehouse CRUD is address forms plus drag-to-reorder, a desktop settings task, and mobile ships no warehouse screen. The test refuses a subtree exemption the moment mobile registers anything under the prefix, so a future read-only mobile list has to be justified route-by-route.

## Tests

**16 new integration tests** — 9 repository, 7 API — including the one the slice exists for: a rename edits the row and the store still has exactly one warehouse afterwards.

Mutation-tested, each mutant confirmed to **compile** first:

| mutation | test that caught it |
|---|---|
| `Create` drops the `ErrNameTaken` mapping | `TestCreate_RejectsADuplicateLiveName` |
| `UpdateByID` drops the archived filter | `TestUpdateByID_RefusesAnArchivedWarehouse` |
| `UpdateByID` drops store scoping | `TestUpdateByID_IgnoresAnotherStoresWarehouse` |
| `SetDefault` loses its transaction | `TestSetDefault_RefusesAnArchivedWarehouse` |
| phone no longer required | `TestAPI_Warehouses_Create_WithoutPhone_400` |
| `Delete` ignores the ownership check | `TestAPI_Warehouses_Delete_AnotherStoresWarehouse_404` |
| `Reorder` accepts a partial set | `TestAPI_Warehouses_Reorder_PartialSet_400` |
| `Delete` reports `archived` without archiving | `TestAPI_Warehouses_Delete_WithStock_ArchivesAndReportsUnits` |

Two mutations were instructive rather than confirmatory and both changed the code:

- **`UpdateByID` without the archived filter initially survived.** The test asserted only the returned error, and the read-back applies the filter — so `ErrNotFound` came back *after* the archived row had already been rewritten. The test now asserts the row's name is untouched.
- **`SetDefault` with the verify moved after the clear also survived**, because the transaction rolls it back. The ordering is not the guard; the transaction is. Dropping the transaction too does fail the test. The comment now says that instead of claiming a reason that was not real.

Full suite, from `services/marketplace-api`, DSN `postgres://dev:dev@192.168.1.110:5432/marketplace_db`:

`go test -tags=integration -count=1 -p 1 ./...` — exit 0, 120 packages ok, **3274 PASS / 25 SKIP / 0 FAIL** (the 25 skips are the pre-existing real-OpenFGA and subscription-trial-cron tests). `go vet -tags=integration ./...` clean.

No migration — 000122 already carries `archived_at` and the partial index — so no `ExpectedSchemaVersion` bump.

## Not in scope

5c (the settings page), 5d (carrier form drops the address; deletes `warehouse-validation.ts`), 5e (per-location stock). Nothing here retires `DefaultLocationID` or touches the allocator — that is PR 6.

### DIFF
--- a/services/marketplace-api/cmd/marketplace-api/main.go
+++ b/services/marketplace-api/cmd/marketplace-api/main.go
@@ -1191,6 +1191,7 @@ func main() {
 			CSVImportsHandler:        csvImportsHandler,
 			PaymentSettingsHandler:   paymentSettingsHandler,
 			ShippingSettingsHandler:  shippingSettingsHandler,
+			WarehousesHandler:        admin.NewWarehousesHandler(conn, log),
 			ShipmentsHandler:         shipmentsHandler,
 			TaxSettingsHandler:       taxSettingsHandler,
 			SettingsMetaHandler:      settingsMetaHandler,
--- a/services/marketplace-api/internal/handlers/admin/errors.go
+++ b/services/marketplace-api/internal/handlers/admin/errors.go
@@ -71,6 +71,9 @@ var codeStatus = map[apperrors.Code]int{
 	apperrors.CodeCampaignNoRecipients: http.StatusUnprocessableEntity,
 	apperrors.CodeCampaignSchedulePast: http.StatusUnprocessableEntity,
 	apperrors.CodeSegmentInUse:         http.StatusConflict,
+
+	// Warehouses — #177 PR 5b.
+	apperrors.CodeWarehouseNameTaken: http.StatusConflict,
 }
 
 // RespondErr writes the standard error envelope for the given error.
--- a/services/marketplace-api/internal/handlers/admin/products_integration_test.go
+++ b/services/marketplace-api/internal/handlers/admin/products_integration_test.go
@@ -138,6 +138,7 @@ func setupTestRouter(t *testing.T) *testEnv {
 		MediaHandler:        mediaHandler,
 		SubscriptionHandler: subHandler,
 		PromoHandler:        promoHandler,
+		WarehousesHandler:   admin.NewWarehousesHandler(db, nil),
 		StoresMiddleware:    storeMW,
 		AuthzMiddleware:     authzMW,
 		InternalSecret:      "",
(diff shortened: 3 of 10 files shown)


mahesh-sangawar:
Superseded by #530 — GitHub auto-closed this when its base branch (`fix/allocator-excludes-archived-warehouses`, merged as #528) was deleted, and a closed PR's base cannot be retargeted. #530 is the same commit rebased onto `main`.

PR #533, opened 2026-09-01
feat(product): per-warehouse stock, end to end (#177 PR 5e)

mahesh-sangawar:
Refs #177. The **backend half** of 5e — the product-form UI is deliberately not here; see *What is not in this PR* below.

## The trap this is built around

`checkout_availability.go` attributes a SENTINEL stock row to the store's **first** warehouse, and *sums* it with any real row for that same warehouse:

```go
avail[r.VariantID][warehouseID] += r.Available
```

Every unit in production is still on the sentinel — I re-checked: **380 rows, 16,354 units, all at `00000000-…-0001`, zero at Main Warehouse.** So a naive per-location write is a stock-doubling bug: a merchant opens a product that reads 10, types 10 against their main warehouse and 5 against a new one, and ends up selling **25**. They changed nothing and doubled their inventory.

`SetVariantStockByLocationInTx` therefore **conserves**: it writes the real rows and drops that variant's sentinel row *in the same transaction*. The edit becomes a per-variant backfill — after it, that variant is fully migrated and PR 6 has one less row to sweep. One transaction because a write that landed without the delete is precisely the double-count.

## Surface

- `PATCH …/variants/:variantId` accepts `inventory_by_location` (warehouse id → quantity). Mutually exclusive with `inventory_quantity`: one names a total with no location, the other locations with no total, and accepting both would leave the service guessing. **400** if both are sent.
- Every named location must be a **live warehouse of this store**. Stock written against another store's warehouse, or an archived one, is stock the allocator will never offer — to the merchant that reads as inventory that vanished.
- `GET …/products/:id` carries `inventory_by_location` per variant. The **sentinel appears under its own id** rather than folded into a warehouse: the admin has to tell "this warehouse holds 10" from "10 units exist but are not assigned anywhere yet", and the merchant's first per-location save is what resolves it. Detail view only — the list renders a total, and a query per page to fill a field nothing reads has no buyer. A failure degrades to the total rather than failing the product load.
- `UpdateVariantStockInTx` remains the single mutation chokepoint; the new method loops through it rather than writing `variant_stock` directly.
- **The single-warehouse path is untouched.** A store with one warehouse sees exactly what it saw before, pinned by a test.

Locations are written in sorted order — map iteration is random, and two concurrent saves touching the same pair in opposite orders can deadlock on the row locks.

## Tests

**11 new integration tests** (6 repository, 5 API + read side), including the one the slice exists for: 10 on the sentinel + a save of A=10, B=5 ends at **15, not 25**.

Six mutations, each confirmed to **compile** first:

| mutation | caught by |
|---|---|
| the sentinel row is left behind — *the double-count bug itself* | 3 repository tests |
| the sentinel accepted as a writable location | *(see below)* |
| an empty map falls through to the delete | `EmptyMapChangesNothing` |
| any warehouse id accepted | 2 API tests |
| both fields accepted together | `BothFieldsIsRefused` |
| per-location writes ignored | `PerLocationSaveConservesTheTotal` |

The sentinel-as-a-location mutation **survived at first** — removing that guard broke no test. `RefusesTheSentinelAsALocation` and `EmptyMapChangesNothing` were written in response, and both now fail without their guards. Worth flagging: two guards I wrote were untested until the mutation pass said so.

Full suite from `services/marketplace-api`, DSN `postgres://dev:dev@192.168.1.110:5432/marketplace_db` — exit 0, 120 packages ok, **3293 PASS / 25 SKIP / 0 FAIL**. `go vet -tags=integration ./...` clean. No migration, so no `ExpectedSchemaVersion` bump.

## What is not in this PR

**The product-form UI.** The rule from the spec — *a store with one warehouse sees exactly what it sees today; the form expands to a per-warehouse breakdown only once a second warehouse exists* — needs warehouses threaded into `ProductForm`, a per-warehouse editor, and the multi-variant tab handled too. That is its own reviewable change, and this contract is worth landing first because it makes the double-count impossible **before** any UI can trigger it.

Nothing here retires `DefaultLocationID` or backfills existing rows — that is PR 6.

### DIFF
--- a/apps/admin/app/(admin)/products/[id]/page.tsx
+++ b/apps/admin/app/(admin)/products/[id]/page.tsx
@@ -7,6 +7,7 @@ import { notFound } from "next/navigation";
 
 import { getServerSessionContext } from "@/lib/auth/serverSession";
 import { getProduct, listCategories } from "@/lib/api/marketplace-api";
+import { listWarehouses } from "@/lib/api/warehouses-api";
 import { Breadcrumbs } from "@/components/layout";
 import { ProductForm } from "@/components/products/ProductForm";
 
@@ -23,9 +24,12 @@ export default async function ProductDetailPage({ params }: PageProps) {
     notFound();
   }
 
-  const [product, categories] = await Promise.all([
+  const [product, categories, warehouses] = await Promise.all([
     getProduct(currentStore.id, id, { userId, tenantId }),
     listCategories(currentStore.id, { userId, tenantId }),
+    // Drives the per-warehouse stock editor (#177 PR 5e). With fewer than
+    // two the form renders exactly as it did before.
+    listWarehouses(currentStore.id, { userId, tenantId }),
   ]);
 
   if (!product) {
@@ -51,6 +55,7 @@ export default async function ProductDetailPage({ params }: PageProps) {
         canArchive={role === "owner" || role === "admin"}
         session={{ userId, tenantId }}
         storeSlug={currentStore.slug}
+        warehouses={warehouses}
       />
     </main>
   );
--- a/apps/admin/app/(admin)/products/actions.ts
+++ b/apps/admin/app/(admin)/products/actions.ts
@@ -24,6 +24,7 @@ import {
   getProduct,
   type CreateProductRequest,
   type UpdateProductRequest,
+  setVariantStockByLocation,
 } from "@/lib/api/marketplace-api";
 import { productFormSchema, type ProductFormValues } from "@/lib/validation/product-form";
 import { bulkActionSchema } from "@/lib/validation/bulk-action";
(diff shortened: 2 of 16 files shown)


mahesh-sangawar:
**Update — the UI half is now on this branch**, so this is the complete 5e slice rather than the backend alone. The PR title and description above cover the contract; this comment covers what the merchant sees.

## What changed in the admin

`GET …/products/:id` already carried `inventory_by_location`; the product edit page now loads the store's warehouses and passes them down.

**The spec's rule, enforced as the visible behaviour:** with fewer than two warehouses the product form renders *exactly* as it did before — the single Stock field, saved through the ordinary product save, untouched. The per-warehouse editor replaces that field only when there is an actual choice to make. Four tests pin it: one warehouse, zero warehouses, and create mode (no saved variant to write against yet) all keep the single field; two warehouses swap it.

**Saving stock is its own action**, not part of the product save. Routing it through the aggregate update would have changed the write path for every merchant with one warehouse, which is the one thing this slice must not do.

**Unassigned units are shown, never guessed.** A variant still on the sentinel gets a banner — *"12 units not yet assigned to a warehouse… the totals replace the unassigned count rather than adding to it"* — and every input starts at **0**. Pre-filling those 12 into the default warehouse would record stock somewhere the merchant never said it was, and it is the merchant's save that resolves the ambiguity (the backend clears the sentinel row in that same transaction).

## Tests

11 new frontend tests. Three mutations, each confirmed to compile first:

| mutation | caught by |
|---|---|
| editor appears with only one warehouse | `one warehouse keeps the single Stock field` |
| sentinel units pre-filled into the default warehouse | `calls out unassigned units without pre-filling them anywhere` |
| negative quantities accepted | `refuses a negative quantity and sends nothing` |

Admin suite: **107 files, 840 tests passing**; `tsc --noEmit` clean. Go suite unchanged from the description above (3293 PASS / 25 SKIP / 0 FAIL).

## Known limitation, deliberately not fixed here

**Multi-variant products keep their per-variant total.** The editor covers the single-variant case; splitting stock per warehouse *per variant* means reworking `VariantMatrixTable`, which is a bigger change than this slice should carry. A merchant with two warehouses and a variant matrix can still set totals exactly as today — they just cannot split them per location yet. Worth its own issue.

Also unchanged: nothing here retires `DefaultLocationID` or backfills the 16,354 existing sentinel units. That is PR 6.

PR #535, opened 2026-09-01
feat(warehouse): backfill sentinel stock onto real warehouses (#177 PR 6, step 1)

mahesh-sangawar:
Refs #177, closes the data half of the contract step. **Migration only — no code changes to the sentinel tolerance, and `DefaultLocationID` is untouched.** See *Why this is only step 1* below.

Every unit in production still sits on `location_id = '00000000-…-0001'`. I re-checked against prod while writing this: **380 rows, 16,354 units, zero at Main Warehouse.** Allocation is structurally correct and economically hollow until this runs.

## The spec is wrong here, and it matters

> production has zero warehouses and zero carrier configs

That was true when the spec was written and false since 2026-09-01. `the-bondi-store` has a warehouse named **exactly `Main Warehouse`**, with a real address, and `warehouses` is keyed `(store_id, name)` — partially, since 000122. The spec has the backfill simply *create* one per store with that name, which against production either violates the index or, with `ON CONFLICT DO NOTHING`, **silently leaves the stock on the sentinel while the code believes it migrated.**

So the migration find-or-creates, and never touches an address a merchant has entered. `TestBackfill_ExistingMainWarehouseIsReusedNotDuplicated` is that case, seeded exactly as bondi is.

## What it does

1. **Creates a warehouse only for stores holding sentinel stock and having no live warehouse at all.** A store that already has one — under *any* name — uses it. The address is left blank deliberately: that store has never said where it ships from, and a plausible-looking placeholder is a lie that quotes rates.
2. **Moves `variant_stock` onto the store's warehouse**, chosen by the same precedence as `warehouse.DefaultForStore` (flagged default, then fill order, then oldest), **excluding archived** — the allocator refuses to fill from those (#528), so backfilling onto one parks every unit somewhere unsellable.
3. **Moves `stock_holds` too.** Deleting them would release a shopper's reservation mid-checkout and let someone else take the units they are paying for.

**Conflicts SUM, they do not overwrite.** A variant can carry both a sentinel row and a real row for the same warehouse — 5e clears the sentinel on edit, but the multi-variant product save still writes it. `checkout_availability` adds them together today, so summing preserves the total the merchant is already selling; overwriting would silently delete stock.

**No temp table.** The first draft materialised the target into `CREATE TEMP TABLE … ON COMMIT DROP`, which only survives if the whole migration runs in one transaction — true under golang-migrate, **false under `psql -f` in autocommit**, which is how a migration gets applied by hand during an incident. It failed exactly that way when I applied it to the dev database. It is a repeated CTE now.

**The down migration refuses to run.** Reversing would mean summing every warehouse row back onto the sentinel, which cannot tell stock this migration moved from a per-warehouse split the merchant made themselves via 5e — a merchant holding 10 in Sydney and 5 in Melbourne would get back 15 in one undifferentiated pile. It raises an exception telling you to restore from a snapshot.

## Why this is only step 1

The spec's expand/contract reasoning still holds: migrations run in the `migrate` initContainer on the **admin** deployment only, while `marketplace-api-storefront` and the sweeper CronJobs roll independently. Removing the sentinel tolerance in the same PR as the backfill risks a storefront pod on the old image reading a location that no longer has rows — every shopper told the item is sold out.

Landing the data move alone is safe in either rollout order: after it there are simply no sentinel rows, and the tolerance never fires. **Step 2** — deleting `DefaultLocationID`, its nine call sites and the tolerance — comes once this is confirmed run in production.

## Tests

**9 integration tests**, executing the migration's own SQL inside a rolled-back transaction rather than asserting against a pre-migrated fixture: the thing that can be wrong here is the SQL, and a test that re-implemented it would only agree with itself.

Four mutations, all caught:

| mutation | caught by |
|---|---|
| no find-or-create guard (the spec's naive INSERT) | 6 tests |
| overwrite instead of summing | `SentinelAndRealRowAreSummedNotOverwritten` |
| archived warehouses accepted as targets | `OlderArchivedWarehouseDoesNotWinOverALiveOne` |
| live holds dropped rather than moved | `LiveHoldsMoveWithTheStock` |

The archived-target mutation **survived the first attempt**: my original test archived the store's only warehouse, so the one the backfill creates is `is_default = true` and wins the ordering regardless. The filter only earns its keep when an archived row would otherwise *sort first* — an older one at a store that already has a live warehouse. That second test is what actually pins it.

## Two guards I tripped

- **`TestEveryIntegrationPackageIsInTheTestIntTarget`** caught that the service ROOT package now has integration tests and `make test-int` never ran it. Added to the Makefile — and the guard's parser only recognised `./`-prefixed patterns, so the root could never be marked covered however it was spelled. Parser extended; removing the entry still fails the guard.
- **`ExpectedSchemaVersion`** bumped to 123.

Applied by hand to the dev database and re-run: idempotent. Full suite — exit 0, 120 packages ok, **3302 PASS / 25 SKIP / 0 FAIL**; `go vet -tags=integration ./...` clean.

## Before merging

This rewrites every stock row in production. Worth taking a Cloud SQL snapshot first — the down migration deliberately will not undo it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,6 +84,7 @@ test-int: ## Run integration tests against the running `make dev` stack
 	@cd services/marketplace-api && \
 	  TEST_DATABASE_URL='postgres://dev:dev@localhost:5432/marketplace_db?sslmode=disable' \
 	  go test -tags=integration -p 1 \
+	    . \
 	    ./cmd/backfill-email/... \
 	    ./internal/apikeys/... \
 	    ./internal/arbitrage/... \
--- a/services/marketplace-api/migrations.go
+++ b/services/marketplace-api/migrations.go
@@ -14,4 +14,4 @@ var MigrationsFS embed.FS
 // migration state doesn't match. M1 scaffold: version 1 is the no-op init
 // migration that seeds the migrations tracking table. Bump this with every
 // real migration added.
-const ExpectedSchemaVersion uint = 122
+const ExpectedSchemaVersion uint = 123
--- a/services/marketplace-api/migrations/000123_backfill_sentinel_stock.down.sql
+++ b/services/marketplace-api/migrations/000123_backfill_sentinel_stock.down.sql
@@ -1,0 +1,16 @@
+-- Deliberately NOT reversible.
+--
+-- Reversing would mean summing every warehouse row back onto the sentinel,
+-- which cannot distinguish stock this migration moved from a per-warehouse
+-- split the merchant made themselves through the product form (#177 PR 5e).
+-- A merchant who genuinely holds 10 in Sydney and 5 in Melbourne would come
+-- back with 15 units in one undifferentiated pile and no record of the
+-- split — silent, and impossible to reconstruct.
+--
+-- Failing loudly is the honest answer. To roll back, restore from a
+-- snapshot taken before the up migration ran.
+DO $$
+BEGIN
+    RAISE EXCEPTION
+        'migration 000123 is not reversible: rolling the sentinel backfill back would merge merchant-entered per-warehouse splits into one pile. Restore from a pre-migration snapshot instead.';
+END $$;
--- a/services/marketplace-api/migrations/000123_backfill_sentinel_stock.up.sql
+++ b/services/marketplace-api/migrations/000123_backfill_sentinel_stock.up.sql
@@ -1,0 +1,139 @@
+-- #177 PR 6 (contract, step 1 of 2) — move every unit off the SENTINEL
+-- location and onto a real warehouse.
+--
+-- variant_stock and stock_holds both carry
+-- location_id = '00000000-0000-0000-0000-000000000001', the DefaultLocationID
+-- constant. Allocation reasons about real warehouse ids, so while stock sits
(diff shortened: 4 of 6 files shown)


mahesh-sangawar:
**Pre-migration backup taken and verified.**

Correcting the PR description: I wrote "Cloud SQL snapshot", which is wrong. Production is **CloudNativePG in-cluster** (`mark8ly-postgres`, 2 instances, primary `mark8ly-postgres-3`), backing up via barman to `gs://tesseract-prod-backups-in/mark8ly-postgres` with a **3-day retention policy**. Scheduled daily at 02:40; the most recent was 8h old.

On-demand backup immediately before this merge:

```
name        mark8ly-postgres-pre-177-pr6-20260901110927
backupId    20260901T110934
phase       completed
started     2026-09-01T11:09:34Z
stopped     2026-09-01T11:09:46Z
error       null
```

**Note the 3-day retention** — this restore point expires 2026-09-04. If the backfill needs reverting after that window, there is nothing to restore from, and the down migration deliberately refuses to run. Worth confirming the result within a day or two rather than leaving it.

PR #543, opened 2026-09-01
refactor(admin): collapse the variant matrix to four columns with a per-row details disclosure

mahesh-sangawar:
Design direction from the `ux-design-specialist` pass on the product page. This is change 1 of 3 from that review; the rail and the create form are not in this PR.

## Before

Eight columns — `Size | Colour | Price | SKU | Stock | Weight | L×W×H | Image` — inside `overflow-x-auto`. Any product with two options scrolled horizontally at a realistic admin width, so Price and Stock (the two things merchants actually scan for) competed for space with fields that are set once and rarely revisited.

## After

Four data columns plus a disclosure:

```
Variant | Price (A$) | SKU | Stock | More ⌄
```

- **One Variant cell.** Every option value composes into a single cell joined by the interpunct — `Bath · Sand`. That separator is already this codebase's idiom for joined metadata (the product header joins handle and storefront link the same way), so it is not a second convention for the same job. Values are emitted in the header's column order.
- **Weight, package dimensions and image move behind a per-row `More` disclosure.** Same borderless ghost-button treatment as the stock-split trigger beside it — nothing draws a box around the trigger — and the same full-width sub-row pattern for the panel, so there is one grammar for row detail rather than two. The panel uses the `border-t border-border-subtle pt-6` hairline that `ProductSection` uses between sections.

## Variants with no options

8 of the-bondi-store's 12 products have variants but no `product_options` rows. Those rows have nothing to compose a name from and previously rendered with **no identity column at all** — `optionNames` derives from `variants[0].optionValues`, which is empty.

They now fall back to their position, in muted ink:

```
Variant 1     189.00    ROBE-S    6    More ⌄
```

Muted rather than full ink is the whole point: it reads as *the system named this row, you didn't* — the same register the Tax summary uses for "the store default applies". Deliberately **not** an em dash (doesn't distinguish row 2 from row 3) and **not** a synthesised "Option A" (that would be fabricating product data).

## Two things that fall out of this

- `VariantImagePicker` loses its card. It was wrapped in `rounded-md border … bg-elevated … shadow-1`, which it needed as a floating popover to separate itself from the row underneath. Inside the details panel there is already a hairline above it, so the wrapper was a bordered card in a system whose rule is hairline rules, not cards.
- `VariantMatrixTable` loses its `md:overflow-visible` exception. That existed *only* so the picker's absolutely-positioned popover could spill out of the row without the wrapper coercing `overflow-y: auto`. The picker no longer escapes the row, so the plain `overflow-x-auto` is correct again.

## Tests

`apps/admin`: **110 files, 869 tests, 0 failures** (863 on `main`; +6 here).

Updated the five tests that asserted the old column layout, and added coverage for the new behaviour:
- composes option values into one cell
- the header is exactly `Variant | Price | SKU | Stock | Details`
- weight and dimensions are absent from the row until expanded, and collapse again
- an option-less variant renders `Variant 3` at index 2
- an option-less variant renders neither a fabricated option name nor an em dash

## Related

- #542 fixes the key collision that made those same option-less rows write edits into each other. Different files, no conflict; either can merge first.
- The `Variant N` fallback is only reachable at all because #540 (now live on `main-0e37834`) stopped the derivation effect emptying those variants on mount.

### DIFF
--- a/apps/admin/components/products/media/MediaCropDialog.test.tsx
+++ b/apps/admin/components/products/media/MediaCropDialog.test.tsx
@@ -72,8 +72,19 @@ describe("MediaCropDialog", () => {
     const onCancel = vi.fn();
     render(<MediaCropDialog sourceUrl="blob:test" onApply={onApply} onCancel={onCancel} />);
     await waitFor(() => expect(screen.getByRole("button", { name: /apply/i })).not.toBeDisabled());
-    fireEvent.keyDown(window, { key: "Enter" });
-    await waitFor(() => expect(onApply).toHaveBeenCalledTimes(1));
+    // The Apply button's disabled state and the window keydown listener are
+    // updated by two different mechanisms: the button is plain render
+    // output, the listener is a passive effect keyed on [pixelCrop]. React
+    // commits the DOM before flushing that effect, so waiting on the button
+    // does not guarantee the listener has re-registered with a non-null
+    // pixelCrop — on a loaded CI runner the keydown lands on the stale
+    // closure and does nothing. Retry the key until it takes; handleApply
+    // sets `busy` synchronously and the handler skips when busy, so a retry
+    // cannot apply twice.
+    await waitFor(() => {
+      fireEvent.keyDown(window, { key: "Enter" });
+      expect(onApply).toHaveBeenCalledTimes(1);
+    });
   });
 
   it("passes sourceMimeType through to cropToBlob", async () => {
--- a/apps/admin/components/products/variants/VariantImagePicker.tsx
+++ b/apps/admin/components/products/variants/VariantImagePicker.tsx
@@ -17,7 +17,12 @@ export function VariantImagePicker({
   const hasMedia = media.length > 0;
 
   return (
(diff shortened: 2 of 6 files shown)


mahesh-sangawar:
Pushed two test fixes on top of the matrix change:

1. **`shows four data columns`** — my own new assertion pinned `Price (USD)`, which is what a full-icu local Node renders; CI's runtime renders `Price ($)`. The test is about the column *set*, so it no longer asserts the ICU-formatted currency symbol.

2. **`MediaCropDialog > triggers Apply on Enter`** — a pre-existing flake, unrelated to this PR, that failed on #542's first run and again here. Root cause: the test waits for the Apply button to become enabled, but the Enter handler is registered by a passive effect keyed on `[pixelCrop]`. React commits the DOM before flushing that effect, so waiting on the button does not guarantee the listener has re-registered with a non-null `pixelCrop` — on a loaded runner the keydown hits the stale closure and no-ops. The keydown is now retried inside `waitFor`; `handleApply` sets `busy` synchronously and the handler skips while busy, so a retry cannot apply twice. Passes 3/3 locally.

Folded in here rather than a separate PR because it was blocking this branch's CI anyway.

Issue #544, opened 2026-09-01
[pricing] Pricing table advertises webhooks and custom code injection — neither is implemented


mahesh-sangawar:
## Problem

The pricing table on `mark8ly.com` advertises two features on paid tiers that do not exist in the product:

| Feature listed | Tier | Reality |
|---|---|---|
| `Read-only API + webhooks` | Studio | The read-only API is real. **Webhooks are not.** |
| `Custom code injection` | Pro | Not implemented. |

Both strings are live right now — confirmed by `curl https://mark8ly.com/` — and originate from `PLAN_META` in `apps/onboarding/components/marketing/Pricing.tsx:70` and `:82`.

## Evidence

**Webhooks** — searched the whole of `services/marketplace-api` for `webhook_subscriptions`, `WebhookSubscription`, `WebhookURL`, `WebhookEndpoint`: zero merchant-facing hits. The two things that look like webhooks are not:
- `internal/webhookevents` — inbound *Stripe* webhook idempotency tracking
- `internal/outbox` — an internal Pub/Sub event-bus pattern

There is no outbound webhook delivery to merchants anywhere in the service. The read-only API half of that bullet *is* real (`plangate.FeatureReadAPI` → `internal/apikeys/*`, `handlers/admin/apikeys_handler.go`).

**Custom code injection** — `plangate.FeatureCustomCodeInjection` exists, but its own comment in `internal/plangate/matrix.go` says it is a placeholder:

> reserved for the future head/body script-injection surface. No handler accepts arbitrary `<script>` payloads today… gate is a forward-compat hedge

So the plan gate is real and the feature behind it is not. A Pro subscriber who goes looking for it finds nothing.

## Why this matters

These aren't roadmap copy on a blog post — they're line items in a pricing table, next to a price, that a merchant pays money against. "Custom code injection" is plausibly a deciding feature for someone choosing Pro over Studio, and "webhooks" is exactly the kind of thing a technical evaluator filters on.

## How this was found

While writing real content for `/help` and `/integrations` (#149), every claim was verified against the source rather than assumed. Three claims failed that check and were dropped from the new pages: webhooks, custom code injection, and PayPal (removed platform-wide by `migrations/000121_retire_untested_paypal`, though `paypal.go` still exists — a good illustration of why "the implementation file exists" is not evidence a feature is offered).

The new pages are now accurate. **The pricing table is not, and it is the higher-traffic surface.**

## Suggested fix

Whichever is true:
[...]

Issue #153, opened 2026-07-01
[marketing] Comprehensive user acquisition & onboarding playbook for Mark8ly


mahesh-sangawar:
This is the single source of truth for how Mark8ly acquires and onboards users. It covers positioning, every traffic channel, the onboarding flow, and email sequences — with checkboxes so progress can be tracked.

---

## 🎯 Positioning & Messaging

Before any channel work, get this locked. Everything flows from it.

**Who we're for:** Independent makers, artisans, and small merchants who want a beautiful storefront without Shopify's fees, plugin sprawl, or the need for a developer.

**The one-liner:** *"A storefront worth opening — quiet commerce for people who actually make things."*

**The three decisive differentiators to lead with in every channel:**
1. 90-day free trial, no card required (competitors offer 3–15 days)
2. 2. Zero platform fee on every sale (Shopify takes 2% unless you use their payments)
3. 3. Designer-quality storefront out of the box — no theme shopping, no consultant
**Tone:** Confident, calm, direct. No hype, no exclamation marks, no "game-changing." Match the product.

---

## 📋 Phase 1 — Foundation (Week 1–2)

Get the basics right before spending time on distribution.

### Website & SEO Foundation
- [x] Fix all Search Console errors (see issues #147, #148, #149, #150)
- [x] Confirm `sitemap.xml` includes all public-facing pages
- [x] Write and publish `/shopify-alternative` landing page (see issue #150)
- [x] Write and publish `/ecommerce-for-makers` landing page (see issue #150)
- [x] Write and publish `/sell-online-india` landing page (see issue #150)
- [x] Write and publish `/etsy-alternative` landing page (see issue #150)
- [x] Add unique `<title>` and `<meta description>` to every page
- [x] Add "Built on Mark8ly" footer link to storefronts (see issue #134)
- [x] Publish first 2 Guides: "How to add your first product" and "How to connect your domain"
- [x] Add an email capture field to the Journal "coming soon" page ("Notify me when the first piece goes up")
### Accounts to Create
- [x] Instagram account — `@mark8ly`
- [ ] LinkedIn page — Mark8ly company page
- [ ] Twitter/X account — `@mark8ly`
- [ ] Pinterest business account
- [ ] Indie Hackers product page at indiehackers.com/product/mark8ly
- [ ] Crunchbase listing
- [ ] Product Hunt upcoming page (to collect followers before launch)
---

## 📣 Phase 2 — Launch Spike (Week 3–4)

[...]

Issue #152, opened 2026-07-01
[seo] Add opt-out "Built on Mark8ly" footer link to merchant storefronts for backlinks


mahesh-sangawar:
## Context

Mark8ly currently has **0 external backlinks**. Domain authority (which drives Google rankings) is primarily built through backlinks from other sites. Every merchant storefront built on Mark8ly is a potential backlink.

## Proposal

Add a small, tasteful "Built on Mark8ly" link in the footer of all merchant storefronts. This should be:

- **Opt-out, not opt-in** — enabled by default, merchants can disable it in their store settings
- - Visually minimal — small text, not a badge or logo, matching the quiet brand aesthetic
- - - A `dofollow` link to `https://mark8ly.com` (this passes SEO authority)
- - - - Example text: `Built on Mark8ly` or `Powered by Mark8ly`
## Implementation Notes

- Add a `showPoweredBy` boolean to merchant/store settings (default: `true`)
- - Render the link in the storefront footer component when `showPoweredBy` is true
- - - The link should use `rel="noopener"` but NOT `rel="nofollow"` (we want the SEO benefit)
- - - - Add the toggle to the admin → Store Settings → Branding section
## Expected Impact

Each live merchant store that Google crawls will pass a small amount of link authority to mark8ly.com. With even 20–50 live stores, this creates a meaningful baseline of domain authority that no other action can replicate as passively.

## Acceptance Criteria

- [ ] Footer link appears on all storefronts by default
- [ ] - [ ] Merchants can disable it in Store Settings → Branding
- [ ] - [ ] Link is `dofollow` (no `rel="nofollow"` attribute)
- [ ] - [ ] Link text and styling is consistent with storefront design aesthetic
- [ ] - [ ] Verified in Search Console Links report within ~4 weeks of first stores going live

mahesh-sangawar:
**Shipped — the dofollow backlink and merchant opt-out are both wired end-to-end.**

Most of this was already built; two gaps against the acceptance criteria are now fixed.

**What already existed**
- The storefront `Footer` renders "Powered by mark8ly" on every page (root `app/layout.tsx`), gated on `branding.show_powered_by` (defaults on → opt-out).
- Backend field `ShowPoweredBy bool` with `default:true` in `marketplace-api` branding models, plus update handling and a `true` default on create.
- An admin toggle under Store Settings → Branding → Advanced.

**What I fixed**
[...]

Issue #151, opened 2026-07-01
[seo] Publish SEO landing pages — Shopify alternative, ecommerce for makers, India UPI


mahesh-sangawar:
## Context

Search Console shows **0 clicks and 0 external backlinks** from any site. The site has only 2 indexed pages and no content targeting search queries. The highest-leverage SEO work right now is building comparison/alternative landing pages that capture high-intent search traffic.

## Pages to Create

These are standalone pages (not blog posts) targeting searches with clear buying intent:

| Page URL | Target Query |
|---|---|
| `/shopify-alternative` | "shopify alternative no transaction fees" |
| `/ecommerce-for-makers` | "ecommerce platform for handmade sellers" |
| `/sell-online-india` | "sell online india UPI payments" |
| `/etsy-alternative` | "etsy alternative own website" |

Each page should be ~800–1200 words, use the existing comparison table data as a foundation, and link back to the homepage and `/onboarding`.

## Why These Pages

- People searching for "Shopify alternative" already have buying intent — they're comparing, not browsing
- - Mark8ly's 90-day free trial + no platform fees is a decisive differentiator to highlight
- - - The India/UPI angle is currently not served by any major competitor well
- - - - These pages will also generate natural backlinks when listed in "best of" roundups
## Acceptance Criteria

- [ ] Each page is live at the specified URL
- [ ] - [ ] Each page is included in `sitemap.xml`
- [ ] - [ ] Each page has a unique `<title>` and `<meta description>` targeting its keyword
- [ ] - [ ] Each page has a clear CTA linking to `/onboarding`
- [ ] - [ ] Pages are confirmed indexed in Search Console within 4 weeks of publish
## Related

See also: Journal and Guides pages (currently "coming soon") — these also need content to contribute to indexable pages.

mahesh-sangawar:
**Shipped — 4 landing pages built and live in the sitemap** (commit `d941f657`, branch `main`).

New routes in the onboarding app (mark8ly.com):
- `/shopify-alternative` → "shopify alternative no transaction fees"
- `/ecommerce-for-makers` → "ecommerce platform for handmade sellers"
- `/sell-online-india` → "sell online india UPI payments"
- `/etsy-alternative` → "etsy alternative own website"

Each page:
- ~800–1200 words, drafted from the existing homepage comparison data and positioning (90-day free trial, no platform fees, UPI/local payments, own domain, one-click export, plans from $29/mo).
[...]

PR #546, opened 2026-09-01
SEO & growth: fix two live 404 bugs, index /help + /integrations, publish guides and Journal capture

mahesh-sangawar:
Works milestone 6 (**SEO & growth**). Audited every open issue against the live site first, which changed the shape of the work: #151 and #152 turned out to be fully built and shipped already, while the audit surfaced two live bugs nobody had a ticket for.

### `www.mark8ly.com` served a soft 404 (`651cda1`)

```
$ curl -s -o /dev/null -w '%{http_code}' https://www.mark8ly.com/
200
$ curl -s https://www.mark8ly.com/ | grep -o '<title>[^<]*</title>'
<title>Store not found</title>
```

**HTTP 200 with an error body**, on the brand's own `www` hostname. `www` routes to the storefront app; `host-policy.ts` classified it as `marketing` correctly, but the middleware comment said *"404 cleanly instead of rendering DEFAULT_STORE under the wrong brand"* while the code beneath it called `passthrough()`. Comment and behaviour had drifted.

It also meant `www.mark8ly.com/robots.txt` served the storefront's robots, advertising `Sitemap: https://default.mark8ly.com/sitemap.xml` — a host whose root is a hard 404.

Now `301`s to the apex of its own TLD. A redirect beats a 404 here: `www` is a real brand alias people type and link, and a 301 consolidates authority onto the canonical apex. Deliberately narrow — the apex itself and platform hosts (`api.`, `admin.`, `uat.`) still pass through so genuine routing drift stays visible, and local dev hosts are never bounced to production.

### Every email address on the site linked to a 404 (`7a4f628`)

Cloudflare's Email Address Obfuscation rewrites `mailto:` into `/cdn-cgi/l/email-protection#...`, and that path 404s here — the Tunnel forwards `/cdn-cgi` straight to the Next.js origin. ~39 anchors across every legal and contact page.

**This was a UX bug before it was an SEO bug:** anyone clicking any email address on the marketing site got a 404 instead of their mail client.

Fixed with a `MailLink` component wrapping each anchor in `<!--email_off-->` / `<!--email_on-->`, the opt-out markers Cloudflare honours. It lives in `@repo/ui` because the bug belongs to the edge rather than any one app — admin's public `/pricing` page had it too. A source-scan test fails the build if a bare `mailto:` anchor is reintroduced.

## Milestone issues

**#149 — `noindex` pages.** The two excluded pages are `/help` and `/integrations`, both thin stubs. Rather than just dropping the tag and handing Google two soft-404 candidates, both got real content first: `/help` is a grouped FAQ with `FAQPage` JSON-LD, `/integrations` is the real integration list traced to the code behind each entry. Both now indexed and in `sitemap.xml`. `/blog`, `/dpa` and the funnel pages stay `noindex`, documented on the issue.

**#153 — two guides + Journal capture.** `how-to-add-your-first-product` and `how-to-connect-your-domain` published, both fact-checked against the actual admin flows. Journal capture is a full vertical slice: a `journal_subscribers` table with **no `tenant_id`** (a subscriber belongs to mark8ly.com, not a merchant), a tenant-free public endpoint that is idempotent and IP-rate-limited and never leaks whether an address was already subscribed, a server-only Next.js proxy, and an accessible form.

**#147, #150** — commented with evidence, left open pending the exact URLs from Search Console. **#151, #152** — verified live and closed.

## Verification

- **Migrations run clean on a fresh database.** Spun a throwaway `postgres:15`, applied all 124 from zero, ran the integration tests against it, then verified the down-migration drops the table and re-up succeeds.
- **Integration tests genuinely pass**, including the double-submit-yields-one-row idempotency test. Worth stating explicitly: they were *failing* when first written because the shared test DB is stuck at schema 119 — this repo's integration tests skip silently and still print `ok` without `TEST_DATABASE_URL`, so "passing" needed proving, not assuming.
- `go build ./...`, `go vet ./...` clean. 54 onboarding unit tests and 81 storefront unit tests pass. `tsc --noEmit` clean for both apps.
- **Not verified: the production build.** `npm install` fails in this worktree on `@tesserix/otto-widget` (GitHub Packages, needs `NODE_AUTH_TOKEN`), so `next build` can't run locally. CI covers it.

## Before merging

⚠️ **`MARKETPLACE_API_URL` must be set on the onboarding deployment** in `tesserix-k8s`, pointing at the marketplace-api **admin** service — the new endpoint is mounted on `mode.Both` and `mode.Admin`, not `mode.Storefront`. Without it the var defaults to `localhost:8088` and the form fails closed with a polite "couldn't save that just now" message. Graceful, but the feature is dead until the env var lands.

Optional: turning off Email Address Obfuscation in Cloudflare → Scrape Shield removes the `/cdn-cgi` root cause outright. The code fix doesn't depend on it and stays correct either way.

## Related

Opened #544 separately — the audit found the **live pricing table advertises two features that don't exist**: "webhooks" on Studio (no outbound webhook delivery exists anywhere in the service) and "custom code injection" on Pro (the plangate flag's own comment calls it a forward-compat hedge with no handler behind it). Both are line items next to a price. Not fixed here because it's a product-copy decision, not SEO work.

Three claims were also cut from the new pages for the same reason — PayPal (removed platform-wide by migration `000121`, though `paypal.go` still exists) and Otto live chat (imported by no page, and its `tenantId` is hardcoded to `"mark8ly"`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,6 +115,7 @@ test-int: ## Run integration tests against the running `make dev` stack
 	    ./internal/handlers/webhooks/... \
 	    ./internal/idempotency/... \
 	    ./internal/inbox/... \
+	    ./internal/journal/... \
 	    ./internal/notification/... \
 	    ./internal/order/... \
 	    ./internal/orderrefund/... \
--- a/apps/admin/app/pricing/layout.tsx
+++ b/apps/admin/app/pricing/layout.tsx
@@ -8,6 +8,8 @@
 import type { ReactNode } from 'react'
 import Link from 'next/link'
 
+import { MailLink } from '@repo/ui/mail-link'
+
 interface PricingLayoutProps {
   children: ReactNode
 }
@@ -48,12 +50,12 @@ export default function PricingLayout({ children }: PricingLayoutProps) {
         style={{ color: 'var(--ink-600)' }}
       >
         <nav aria-label="Footer" className="flex gap-6 text-sm">
-          <a
-            href="mailto:hello@mark8ly.com"
+          <MailLink
+            email="hello@mark8ly.com"
             className="hover:text-[var(--moss-700)] transition-colors focus:outline-none focus-visible:ring-2 focus-visible:ring-[var(--moss-700)] rounded-sm"
           >
             Contact
-          </a>
+          </MailLink>
           <Link
             href="/privacy"
             className="hover:text-[var(--moss-700)] transition-colors focus:outline-none focus-visible:ring-2 focus-visible:ring-[var(--moss-700)] rounded-sm"
--- a/apps/onboarding/.env.example
+++ b/apps/onboarding/.env.example
@@ -5,6 +5,11 @@ PORT=4201
 # These never reach the browser.
 PLATFORM_API_URL=http://localhost:8086
 AUTH_BFF_URL=http://localhost:8087
+# marketplace-api base URL — used by the /api/journal-subscribe route only
+# (issue #153, the Journal "coming soon" page). Not PLATFORM_API_URL: that
+# points at platform-api, a different service. Matches the env var name
+# and default port apps/admin/lib/api/marketplace-api.ts already uses.
+MARKETPLACE_API_URL=http://localhost:8088
 
 # Browser-exposed (NEXT_PUBLIC_*).
 # All three are public by design — they identify the GIP project but
--- a/apps/onboarding/app/acceptable-use/page.tsx
+++ b/apps/onboarding/app/acceptable-use/page.tsx
@@ -1,3 +1,4 @@
+import { MailLink } from "@repo/ui/mail-link";
 import {
   MarketingPage,
   PageHero,
@@ -72,7 +73,7 @@ export default function AcceptableUsePage() {
         <p>
           If your account is suspended or terminated and you believe we got
           it wrong, email{" "}
-          <a href="mailto:legal@mark8ly.com">legal@mark8ly.com</a> within 30
+          <MailLink email="legal@mark8ly.com" /> within 30
           days. Include your store name, the action we took, and why you think
           it was incorrect. We respond within 10 business days.
         </p>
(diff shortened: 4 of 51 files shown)


mahesh-sangawar:
## CI caught a real bug in this branch — a privacy one

`TestErasurePlan_CoversEveryCustomerLinkedTable` failed:

```
--- FAIL: TestErasurePlan_CoversEveryCustomerLinkedTable
    coverage_integration_test.go:66
    Error: Should be empty, but was [journal_subscribers]
```

The guard asserts that every table carrying a customer-link column (here, `email`) either has a step in the erasure plan or a *justified* entry in `declaredExclusions`. The new `journal_subscribers` table had neither.

**My mistake:** I ran only `./internal/journal/...` and `./internal/handlers/public/...` locally rather than the full integration suite, so I never saw this. Fixed by running the whole suite the way CI does — it now passes end to end with zero failures.

### Disposition: declared exclusion, with reasoning

`journal_subscribers` is deliberately excluded rather than erased, because a store-scoped erasure plan **structurally cannot reach it**: the table carries no `store_id` and no `tenant_id`, since a subscriber belongs to mark8ly.com rather than to any merchant's store. There is no key to erase by, and erasing on a customer's request would delete a platform subscription that person may never have made.

The justification says so explicitly, and is careful to state what it does *not* mean:

> NOT a statement that these addresses are unerasable: they are personal data and still carry an art.17 right, exercised against the platform rather than through a merchant's store-scoped flow.

Fixed in `7b1b8a9`.

### Follow-up worth tracking separately

That justification is honest about the right existing — but **there is currently no mechanism to exercise it.** `journal_subscribers` has no unsubscribe link and no platform-level erasure path. That's acceptable while the Journal has no articles and no one is subscribed, and it should not stay that way once the first piece ships and the form starts collecting addresses.

Not fixed here — it's a separate piece of work, and pretending otherwise inside an SEO PR would bury it. Happy to open an issue for it if that's wanted.

---

Worth saying plainly: this guard did its job. A new table holding email addresses tried to enter the schema without anyone deciding what happens to that data on an erasure request, and the build stopped it.

PR #555, opened 2026-09-02
chore(deps): move @tesserix packages to the public npm registry

mahesh-sangawar:
Completes the migration `CLAUDE.md` has been flagging: *"consumers still pinning `@tesserix:registry=https://npm.pkg.github.com` with `PKG_READ_TOKEN` must move to the public registry."*

## Why now

I went to bump `@tesserix/web` so the AlertDialog fix (2.4.1, published today) would reach admin, and found mark8ly **cannot see any 2.x at all**. Three `.npmrc` files still route the scope to GitHub Packages:

```
@tesserix:registry=https://npm.pkg.github.com
//npm.pkg.github.com/:_authToken=${NODE_AUTH_TOKEN}
```

The design system publishes to public npm now. GitHub Packages' last `@tesserix/web` was **1.7.1**, which is exactly where mark8ly has been frozen.

## What changed

- **Deleted all three `.npmrc` files.** They contained nothing but the scope override and its token.
- **`@tesserix/otto-widget` `^0.5.4` → `^0.6.0`.** Not optional: public npm has `0.4.2 … 0.5.1, 0.6.0` — **`0.5.4` never existed there**, it was a GitHub-Packages-only build, so `^0.5.4` is unsatisfiable post-migration.
- **`@tesserix/web` stays `^1.7.1`**, now resolving to **1.8.1** from public npm. See below for why it does not go to 2.x.
- **Dockerfiles**: dropped `.npmrc` from the `COPY` line — *this would otherwise have failed the build outright* — and removed the `PACKAGE_READ_TOKEN` BuildKit secret and `NODE_AUTH_TOKEN`, which nothing reads any more.

Every `@tesserix/*` entry now resolves from `registry.npmjs.org`; `grep -c 'npm.pkg.github.com' package-lock.json` → **0**.

## Why @tesserix/web stops at 1.8.1

I tried 2.4.1. It installs, and then 9 admin test files fail to load:

```
SyntaxError: Named export 'Github' not found.
The requested module 'lucide-react' is a CommonJS module…
```

`@tesserix/web@2.4.1` declares `peerDependencies: { "lucide-react": "^0.469.0" }`. **mark8ly is on lucide-react 1.14.0**, and the 0.x → 1.x release removed the brand icons (`Github`, `Gitlab`, …) that the design system's auth provider marks import. `--legacy-peer-deps` lets the mismatch install silently, and it only surfaces at module evaluation.

That is a design-system problem, not one mark8ly can fix from this side — it needs to support lucide-react 1.x, or stop importing icons lucide has removed. Filed separately. Notably 2.0.0's documented breaking change is confined to the auth surface (`Aurora*` → `Auth*`), which mark8ly does not use at all, so once the peer range is sorted the upgrade should be close to mechanical.

**Consequence worth stating plainly:** the AlertDialog portal fix and the oxblood confirm button do **not** reach production with this PR. They are blocked behind the lucide incompatibility.

## Unexpected win

Admin's typecheck is now **completely clean**. It had two standing errors:

```
components/support/LiveChatInbox.tsx(4,27): error TS2307: Cannot find module '@tesserix/otto-widget'
components/support/PlatformChat.tsx(3,28): error TS2307: Cannot find module '@tesserix/otto-widget'
```

`otto-widget@0.6.0` from public npm ships types the GitHub Packages 0.5.4 build did not. I had previously written these off as unrelated noise; they were a symptom of the same registry split.

## Verification

| suite | result |
|---|---|
| `apps/admin` | 111 files, **884 tests**, 0 failures |
| `apps/storefront` | 10 files, **81 tests**, 0 failures |
| `apps/onboarding` | **68 tests**, 0 failures |
| `apps/admin` typecheck | clean (was 2 errors) |
| `apps/onboarding` typecheck | clean |
| `apps/storefront` typecheck | clean apart from a stale `.next/types/validator.ts` referencing a since-renamed route — a gitignored build artifact, not from this change |

## Note for reviewers

The shared `container-ci.yml` / `container-release.yml` workflows still pass `package_read_token`. That is harmless — an unused secret — but it can be dropped from `tesserix-workflows` whenever convenient, and no longer needs to be kept valid for mark8ly's sake.

### DIFF
--- a/.github/scripts/test-ci-contract.py
+++ b/.github/scripts/test-ci-contract.py
@@ -159,8 +159,24 @@ class ReusableCIContract(unittest.TestCase):
         for app, google_slot in next_apps.items():
             contents = (ROOT / f"apps/{app}/Dockerfile").read_text()
             with self.subTest(app=app):
-                self.assertIn("id=PACKAGE_READ_TOKEN", contents)
+                # PACKAGE_READ_TOKEN is deliberately NOT required. Every
+                # @tesserix/* package now comes from the public npm registry,
+                # so `npm ci` needs no credential and the scoped-registry
+                # .npmrc is gone. The reusable workflow still passes the
+                # secret; nothing consumes it. Asserting its ABSENCE instead
+                # would be wrong too — a Dockerfile is free to mount secrets
+                # this contract does not know about.
                 self.assertIn("id=APPLICATION_BUILD_SECRET", contents)
+                # The registry move is only complete if no .npmrc is COPYed
+                # in: the file no longer exists, so that instruction would
+                # fail the build outright. Checked against COPY lines rather
+                # than the whole file, so a comment explaining the migration
+                # does not trip it.
+                copied = [
+                    line for line in contents.splitlines()
+                    if line.startswith("COPY") and ".npmrc" in line
+                ]
+                self.assertEqual(copied, [], "Dockerfile still COPYs a deleted .npmrc")
                 self.assertIn("ARG REUSABLE_BUILD_CACHE_FP", contents)
                 self.assertIn(
                     f"NEXT_PUBLIC_GOOGLE_CLIENT_ID=${google_slot}", contents
--- a/.github/scripts/test-ci-contract.py
+++ b/.github/scripts/test-ci-contract.py
@@ -177,6 +177,12 @@ class ReusableCIContract(unittest.TestCase):
                     if line.startswith("COPY") and ".npmrc" in line
                 ]
                 self.assertEqual(copied, [], "Dockerfile still COPYs a deleted .npmrc")
+                # The builder must carry the per-workspace node_modules, not
+                # just the root tree. npm does not always hoist a workspace
+                # dependency, and copying only /src/node_modules drops it —
+                # which fails as "Module not found" during next build, long
+                # after the install step reported success.
+                self.assertIn("COPY --from=deps /src/apps ./apps", contents)
                 self.assertIn("ARG REUSABLE_BUILD_CACHE_FP", contents)
                 self.assertIn(
                     f"NEXT_PUBLIC_GOOGLE_CLIENT_ID=${google_slot}", contents
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,0 @@
-@tesserix:registry=https://npm.pkg.github.com
-//npm.pkg.github.com/:_authToken=${NODE_AUTH_TOKEN}
--- a/apps/admin/Dockerfile
+++ b/apps/admin/Dockerfile
@@ -33,6 +33,14 @@ FROM ghcr.io/tesserix/base-node-builder-22@sha256:517372359a366695db0898e2295158
 USER 0:0
 WORKDIR /src
 COPY --from=deps /src/node_modules ./node_modules
+# npm does not always hoist a workspace dependency to the root. When two
+# workspaces resolve a package the root tree cannot satisfy, it lands in
+# apps/<app>/node_modules instead — and copying only /src/node_modules
+# silently drops it, which surfaces much later as "Module not found" during
+# next build rather than as an install error. Carry the per-workspace trees
+# too; the context excludes node_modules, so the COPY below cannot clobber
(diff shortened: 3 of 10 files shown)


mahesh-sangawar:
CI caught something I had missed, and it was a real gap rather than a flake — it failed identically on both attempts.

`Product policy` runs `.github/scripts/test-ci-contract.py`, which asserted every Next.js Dockerfile mounts the package-read secret:

```python
self.assertIn("id=PACKAGE_READ_TOKEN", contents)
```

That was the contract while `@tesserix/*` lived on GitHub Packages. It is stale now, so the assertion is removed — with a comment saying why, and explicitly noting that asserting its *absence* would be wrong too, since a Dockerfile is free to mount secrets this contract knows nothing about.

I replaced it with the check this migration actually needs:

```python
copied = [line for line in contents.splitlines()
          if line.startswith("COPY") and ".npmrc" in line]
self.assertEqual(copied, [], "Dockerfile still COPYs a deleted .npmrc")
```

`COPY … .npmrc ./` against a deleted file fails the build outright, and that is precisely the mistake this PR had to fix in all three Dockerfiles. Now it cannot come back silently.

Two notes on getting there:

- My first version of that assertion was `assertNotIn(".npmrc", contents)`, which failed against my own explanatory comment mentioning `.npmrc`. Scanning `COPY` lines is the precise check.
- Mutation-verified: restoring `.npmrc` to admin's `COPY` line fails with `Lists differ: ['COPY package.json package-lock.json .npmrc ./'] != []`, and removing it passes 6/6.

mahesh-sangawar:
All three Next.js container builds failed — twice, identically, so I reproduced it locally with `docker build` rather than re-running CI again.

## What failed

```
Turbopack build failed with 49 errors:
Module not found: Can't resolve '@tesserix/web'
```

`npm ci` itself **succeeded** — 1036 packages, versus 1034 on the last green build. So the install was fine; the package was missing at build time.

## Why

`npm ci` in the image does install `@tesserix/web@1.8.1` — but at **`apps/<app>/node_modules/`**, not hoisted to the root. Confirmed inside the image:

```
$ docker run --rm … npm ls @tesserix/web
mark8ly@ /src
+-- @mark8ly/admin@0.0.1 -> ./apps/admin
| `-- @tesserix/web@1.8.1
+-- @mark8ly/onboarding@0.0.1 -> ./apps/onboarding
| `-- @tesserix/web@1.8.1
`-- @mark8ly/storefront@0.0.1 -> ./apps/storefront
  `-- @tesserix/web@1.8.1
```

and the lockfile agrees — there is no root entry at all, only three per-workspace ones.

The builder stage copied only the root tree:

```dockerfile
COPY --from=deps /src/node_modules ./node_modules
```

so everything under `apps/*/node_modules` was silently dropped. The install reported success and the failure surfaced much later as `Module not found` during `next build`.

## Why it only bit now

This is a **latent fragility, not something this PR introduced**. While `@tesserix/web` came from GitHub Packages there was exactly one resolvable version and npm hoisted it to the root, so copying the root tree happened to be sufficient. Moving to public npm let the three apps resolve 1.8.1, npm placed it per-workspace, and the gap opened. Any future per-workspace install would have done the same.

## Fix

```dockerfile
COPY --from=deps /src/node_modules ./node_modules
COPY --from=deps /src/apps ./apps
COPY . .
```

Safe ordering: the build context excludes `node_modules` via `.dockerignore`, so the following `COPY . .` cannot clobber what was just copied.

**Verified by building the real image locally**, not by reasoning:

```
$ docker build --target builder -f apps/admin/Dockerfile .   → DONE
$ docker run --rm … ls apps/admin/node_modules/@tesserix/    → web
$ docker run --rm … ls apps/admin/.next/standalone/apps/admin/server.js
                                                             → server.js
```

The contract test now asserts the copy is present, with a comment explaining the failure mode, so this cannot regress quietly.

## On my process

I validated this PR with `npm install` and the app test suites, and never ran `npm ci` or a container build — which is exactly what CI does and exactly where it broke. The lockfile inconsistency was visible the whole time; I just never looked at it the way the builder does.

Issue #561, opened 2026-09-02
The npm audit gate fails on a mobile-only dependency, blocking every PR — and image-size has no patched version


mahesh-sangawar:
`Next.js / Next.js quality` fails at **Audit production dependencies** on every PR right now. Tests pass (95 / 81 / 908); only the audit step fails. It is currently blocking #559, which carries the AlertDialog fix.

## The advisory has no fix

```
image-size  *
Severity: high
  GHSA-w3rx-r6r6-pgpr — ICNS parser DoS via infinite loop
  GHSA-5p2g-fcmc-qvqq — JXL and HEIF parsers DoS via infinite loops
```

From the advisory data:

```
$ gh api /advisories/GHSA-w3rx-r6r6-pgpr
affected: "<= 2.0.2"    first_patched_version: null
```

Latest published `image-size` is **2.0.2**. **Every released version is affected**, so the root-`overrides` approach that cleared the browserslist advisory (#545) has nothing to point at. `npm audit fix --force` proposes `react-native@0.86.3`, a breaking change that would not help either.

## It is not reachable from any audited workspace

```
$ npm ls image-size
mark8ly@
└─┬ @repo/mobile-admin@0.0.0 -> ./apps/mobile-admin
  ├─┬ expo@56.0.15 → @expo/metro → metro@0.84.4 → image-size@1.2.1
  └─┬ react-native@0.85.3 → @react-native/community-cli-plugin → metro
```

It arrives solely through `apps/mobile-admin`, as React Native / Metro **bundler tooling**. `mobile-admin` is deliberately not in `audit_workspaces`:

```yaml
audit_workspaces: >-
  @mark8ly/admin @mark8ly/storefront @mark8ly/onboarding @repo/ui @repo/otto-widget
```

But the step runs:

```bash
npm audit --audit-level=high --omit=dev --workspace=@mark8ly/admin …
```

and `npm audit --workspace=…` still walks the **hoisted root tree**, so it reports packages no audited workspace depends on. The gate is over-reporting against its own stated intent.

## Suggested fix, in `tesserix-workflows`

Fail only when a vulnerable package is actually reachable from an audited workspace — audit as JSON and filter on the dependency paths:

```bash
npm audit --json --omit=dev > audit.json || true
jq --argjson ws "$(printf '%s\n' $NPM_AUDIT_WORKSPACES | jq -R . | jq -s .)" '
  [ .vulnerabilities[]
    | select(.severity == "high" or .severity == "critical")
[...]

PR #568, opened 2026-09-02
chore(ci): allowlist the two unfixable image-size advisories so the audit gate stops blocking every PR

mahesh-sangawar:
Unblocks #561. Consumes `tesserix-workflows` **v2.3.0**, which adds the optional `audit_allowlist` input.

## Why the merge upstream was not enough

mark8ly pins the reusable workflows by tag. Merging the fix in `tesserix-workflows` changed nothing here while `ci.yml` still said `@v2.2.1` — the same shape of silent non-effect as the rest of #561. All five pins move to `v2.3.0`; the only file that changed between the tags is `nextjs-ci.yml`, so the other four are a no-op and move together for consistency.

## The allowlist

```yaml
audit_allowlist: >-
  GHSA-w3rx-r6r6-pgpr
  GHSA-5p2g-fcmc-qvqq
```

Both are `image-size`, and both are **unfixable**: the advisory range is `<= 2.0.2` with `first_patched_version: null`, and 2.0.2 is the newest release. There is no version to upgrade to and no override to write.

It reaches this tree only through the mobile apps — `react-native`/`expo` → `metro` → `image-size` — as bundler tooling. It is in no web workspace's dependency graph and ships in no web image. `audit_workspaces` already says that, but `--workspace` does not scope `npm audit`; verified twice:

```
$ npm audit --audit-level=high --omit=dev --workspace=@mark8ly/admin …   → 6 high
$ cd apps/admin && npm audit --audit-level=high --omit=dev               → 6 high
```

which is why an explicit allowlist is needed rather than a narrower workspace list.

## Only these two, and nothing else, is being accepted

Six packages are reported — `image-size`, `metro`, `metro-config`, `metro-transform-worker`, `react-native`, `@react-native/community-cli-plugin`. I replicated the gate's rule locally (a vulnerability passes only when **every** advisory it stems from is allowlisted) and walked each one's `via` chain:

```
@react-native/community-cli-plugin  stems from [GHSA-5p2g-fcmc-qvqq, GHSA-w3rx-r6r6-pgpr] -> tolerated
image-size                          stems from [GHSA-5p2g-fcmc-qvqq, GHSA-w3rx-r6r6-pgpr] -> tolerated
metro                               stems from [GHSA-5p2g-fcmc-qvqq, GHSA-w3rx-r6r6-pgpr] -> tolerated
metro-config                        stems from [GHSA-5p2g-fcmc-qvqq, GHSA-w3rx-r6r6-pgpr] -> tolerated
metro-transform-worker              stems from [GHSA-5p2g-fcmc-qvqq, GHSA-w3rx-r6r6-pgpr] -> tolerated
react-native                        stems from [GHSA-5p2g-fcmc-qvqq, GHSA-w3rx-r6r6-pgpr] -> tolerated

predicted gate result: PASS
```

All six trace back to exactly these two advisories. If any of those packages later picks up an *unrelated* advisory, the gate fails again — which is the property the upstream input was designed around, and the reason this is an allowlist of advisory ids rather than package names.

The gate keeps its teeth: it would still have caught the browserslist advisory that blocked everything this morning (#545), which **is** reachable from the web workspaces.

## Why this matters beyond merges

The audit gate blocks **deploys**, not just PRs. `Images` declares `needs: [policy, go, nextjs, secrets]`, so a red `nextjs` skips the container build entirely:

```
failure  Next.js / Next.js quality
skipped  Images
skipped  Containers
```

Three commits are on `main` and not in production behind this, including #559 (`@tesserix/web@2.4.2` — the AlertDialog portal fix and the destructive-coloured confirm button). Admin-merging past the gate got code onto `main` but never into a running pod.

## Removal condition

Delete both ids the moment a patched `image-size` ships. The comment in `ci.yml` says so, next to the reasoning, so the next person does not have to reconstruct why they are there.

### DIFF
--- a/.github/scripts/test-ci-contract.py
+++ b/.github/scripts/test-ci-contract.py
@@ -10,7 +10,10 @@ ROOT = Path(__file__).resolve().parents[2]
 WORKFLOW = ROOT / ".github/workflows/ci.yml"
 IMAGES = ROOT / ".github/ci/container-images.json"
 CANDIDATE_REF = "29f963da2412a4ba0c755f19697ad0a31d7624b4"
-RELEASE_REF = "v2.2.1"
+# Bumped to v2.3.0 for the `audit_allowlist` input on nextjs-ci.yml (#561).
+# That tag's only change from v2.2.1 is nextjs-ci.yml, so the other four
+# callers move with it purely to keep one ref across the file.
+RELEASE_REF = "v2.3.0"
 
 
 class ReusableCIContract(unittest.TestCase):
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,7 @@ jobs:
           - service: otto
             directory: services/otto
             coverage: 4
-    uses: tesserix/tesserix-workflows/.github/workflows/go-ci.yml@v2.2.1
+    uses: tesserix/tesserix-workflows/.github/workflows/go-ci.yml@v2.3.0
     with:
       working_directory: ${{ matrix.directory }}
       go_version: 1.26.6
@@ -67,7 +67,7 @@ jobs:
 
   nextjs:
     name: Next.js
-    uses: tesserix/tesserix-workflows/.github/workflows/nextjs-ci.yml@v2.2.1
+    uses: tesserix/tesserix-workflows/.github/workflows/nextjs-ci.yml@v2.3.0
     with:
       working_directory: .
       node_version: "22"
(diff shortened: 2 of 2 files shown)


mahesh-sangawar:
## The allowlist works — `Next.js quality` now passes

First run confirmed the fix: `Audit production dependencies` is no longer in the failure list, so the two `image-size` ids cover all six reported packages exactly as predicted locally.

What failed instead was **`Product policy`**, mark8ly's own contract test:

```
FAIL: test_caller_is_thin_explicit_and_fail_closed
AssertionError: 'v2.3.0' not found in {'29f963da…', 'v2.2.1'}
```

`.github/scripts/test-ci-contract.py` pins the reusable-workflow refs the caller may use, and `RELEASE_REF` was still `v2.2.1`. That is the guard doing its job — it exists so a caller cannot drift onto an unreviewed ref — so the fix is to move the pin deliberately, not to loosen the check.

`RELEASE_REF = "v2.3.0"`, with a comment recording why it moved and that v2.3.0's only change from v2.2.1 is `nextjs-ci.yml`.

Verified the guard still bites: leaving a single caller on `@v2.2.1` fails with `AssertionError: 1 != 2`, and with all five aligned it passes 6/6.

mahesh-sangawar:
Reopening to trigger CI — GitHub did not create a workflow run for e3f5218a.

mahesh-sangawar:
Superseded by #567, which landed the identical change while this was in CI: the same two `image-size` advisory ids, all five callers on `@v2.3.0`, and `RELEASE_REF = "v2.3.0"` in the contract test.

Closing rather than resolving the conflict — there is nothing here that #567 does not already have on `main`, and the verification is worth keeping only as a record: the allowlist was confirmed sufficient and minimal by replicating the gate's rule locally (a vulnerability passes only when *every* advisory it stems from is listed), which showed all six reported packages tracing back to exactly those two ids.

PR #666, opened 2026-09-04
fix(ci): repin base images to the 20260904 weekly rebuild

mahesh-sangawar:
Every container job in the repo is failing. This unblocks them.

## What happened

The weekly base-image rebuild ran at 05:40 today and published new digests tagged `20260904`/`latest`/`weekly`, pruning the previous versions. Every Dockerfile pins its base by digest, so all four pinned images went missing at once:

```
ghcr.io/tesserix/base-distroless-static@sha256:0d68ca80… : not found
ghcr.io/tesserix/base-go-builder@sha256:4ec5ce58…       : not found
```

Builds fail at `[internal] load metadata`, before any repo code is compiled — which is why the failure is identical across `auth-bff`, `platform-api`, `marketplace-api`, `mcp`, `otto`, `admin`, `storefront` and `onboarding`, regardless of what the PR changed.

## The fix

Repins all four base images across 8 Dockerfiles to the current `latest` digests, each verified present in the registry:

| Image | New digest |
|---|---|
| `base-distroless-static` | `sha256:db09c7f5…a99598` |
| `base-go-builder` | `sha256:1d52f132…dd7d1f` |
| `base-node-builder-22` | `sha256:a86cc601…17a855` |
| `base-node-runtime-22` | `sha256:0e94a9fe…62cfe43` |

## Follow-up worth having

Nothing under `.github/` references these images, so the repin after each weekly rebuild is a manual step that nobody is prompted to do — the failure surfaces later as every PR breaking at once, with an error that looks unrelated to the PR. Either the rebuild workflow should open this PR automatically, or the old digests should be retained long enough for a repin to land.

### DIFF
--- a/apps/admin/Dockerfile
+++ b/apps/admin/Dockerfile
@@ -8,7 +8,7 @@
 # differ.
 
 # ─── deps ──────────────────────────────────────────────────────────────
-FROM ghcr.io/tesserix/base-node-builder-22@sha256:517372359a366695db0898e2295158cf4b227537377a264dc55a7faa623cfc29 AS deps
+FROM ghcr.io/tesserix/base-node-builder-22@sha256:a86cc6019a5c1fd8601c68fe076a178aef575bc40f38ade51862efb48017a855 AS deps
 USER 0:0
 WORKDIR /src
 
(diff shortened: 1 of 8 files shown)


mahesh-sangawar:
Superseded by #665, which landed the same repin first — I verified the digests are byte-identical to this branch's for every base image. Closing as redundant.

The follow-up in this PR's description still stands and isn't addressed by #665: nothing under `.github/` references these base images, so the repin after each weekly rebuild is a manual step nobody is prompted to do. It surfaces as every PR in the repo failing at once, in ~20s, with a `load metadata ... not found` error that looks unrelated to the change. Either the rebuild workflow should open the repin PR itself, or old digests need a retention window long enough for one to land.

Issue #412, opened 2026-08-28
MCP: the storefront OpenAPI spec is never ingested — wrong port and path, silently


mahesh-sangawar:
## The existing MCP integration has never worked

`internal/handlers/storefront/openapi.go` describes itself as "the Phase 3 piece of the SLM/MCP coverage work" and defines five `customer-read` operations for `mcp-gateway` to auto-register. It is correct, served, and covered by tests.

**None of its operations are registered as tools.** Asked directly, the gateway returns 8 tools and not one of them comes from the spec:

```
tools/list ->  search_knowledge_base, lookup_conversation, get_order, list_returns,
               check_payment_status, list_recent_orders, create_refund_request,
               create_support_ticket
```

All 8 are gateway-native. The spec's five —

```
getStoreBranding, listStoreCategories, listProductsByCategory,
listStoreProducts, getStoreProduct
```

— are absent. So an agent cannot browse a store's catalogue, categories or branding: exactly what the spec was written to enable.

## Cause: the configured URL is wrong twice

`mark8ly-mcp` (ArgoCD app, `charts/apps/mcp-gateway`) is configured with:

```
openapi.urls = http://mark8ly-marketplace-api-storefront.mark8ly.svc.cluster.local/storefront/openapi.json
```

1. **No port.** It defaults to `:80`; the Service only exposes `:8080` (`port=8080 -> targetPort=http`).
2. **Missing the `/api/v1` prefix.** `ServeOpenAPISpec` is mounted on the `/api/v1` group.

Verified in-cluster against the storefront pod:

```
:8080/storefront/openapi.json          -> 404
:8080/api/v1/storefront/openapi.json   -> 200   (valid OpenAPI 3.0, 5 operations)
```

## Fix

```
openapi.urls = http://mark8ly-marketplace-api-storefront.mark8ly.svc.cluster.local:8080/api/v1/storefront/openapi.json
```

A one-parameter change in `argocd/prod/apps/ai-apps/mark8ly-mcp.yaml` (`tesserix-k8s`).

## Why it went unnoticed, and what to add with the fix

The failure is silent by construction: `mcp-gateway` stays `Healthy`, `/readyz` returns `{"ok":true,"tenant":"mark8ly"}`, ArgoCD reports `Synced`, and the pod serves its native tools normally. Nothing anywhere reports that a spec fetch 404'd. Same shape as #280 — built, tested, deployed, unreachable — and #407, where every surface said success while the thing itself never happened.

[...]

PR #676, opened 2026-09-04
docs(mcp-catalog): the plan missed AgentGateway, and had the smoke port wrong

mahesh-sangawar:
Two corrections to the deployment plan merged in #675, both found by checking work that already exists rather than by reasoning from the plan.

## 1. The plan had no AgentGateway tasks, and needed three

Kora shipped this exact problem days ago. `kora/docs/architecture/agentic-ai-end-to-end.md` (verified 2026-09-04) establishes the sanctioned call path, and it is not what #675 assumed:

**The agent never holds the MCP key.** It presents a short-lived Zitadel JWT to AgentGateway, which verifies signature, issuer, audience, expiry and an `agentgateway.mcp` role in strict mode, and only then injects `X-MCP-Key` on the way to the connector. The model never sees that value.

That injection reads from `product-mcp-upstream-keys` in `agentgateway-system` — built by `external-secrets/prod/agentgateway-system/externalsecret.yaml`, already holding `HOMECHEF_MCP_KEY`, `MARK8LY_MCP_KEY`, `PLATFORM_MCP_KEY` and `STOCKPILOT_MCP_KEY`, and it is exactly the `credentialRef.secretName` devai's registry seeds point at.

So the connector's key has **two readers, not one**. #675 wired only the connector's own copy, which would have produced a deployment that passes every health check and rejects every call the gateway makes.

Three changes:

- **Task 2** now states the two-reader fact up front, so the operator minting the key knows it is not just for the connector.
- **New Task 3b** — add the key to `product-mcp-upstream-keys`, and verify the sync by printing key NAMES only, never values.
- **New Task 3c** — admit the gateway at the mesh boundary. Kora describes two: the waypoint authorizes the original `agentgateway-mcp` or `support-platform-slm-router` caller at the Service boundary, and the workload boundary admits those or the waypoint's forwarding identity, over ambient mTLS by SPIFFE identity. The task says to read how `kora-mcp` actually does it — including `charts/thirdparty/istio-config/templates/network-policies.yaml`, which names it — rather than infer.

Open question 2 ("does the engine reach it directly or via AgentGateway?") is now **answered: via AgentGateway**, and that answer is what produced these tasks.

A new open question 3 replaces it, honestly unresolved: **how is an MCP server registered as an AgentGateway route?** `charts/thirdparty/agentgateway/values.yaml`'s `backends` block is for LLM providers, not MCP servers, so routing lives elsewhere and I have not found it. The plan now says to establish that before assuming a deployed Service is reachable — a connector the gateway cannot route to is running but unreachable, which is #412's shape a second time.

## 2. The plan's `smoke_port` was wrong, and would have reddened `main`

#675 specified `"smoke_port": 8765`. Review of the Task 1 implementation caught it: `config.Load()` refuses to start without `STOREFRONT_BASE_URL`, `STOREFRONT_KEY` and `MCP_AUTH_KEY`, and `main.go` calls it before anything binds a port — while CI supplies only registry tokens and `NEXT_PUBLIC_*` build args.

The convention was already in the file. **Every** Go service uses `smoke_port: 0`:

| entry | port |
| --- | --- |
| `mark8ly-platform-api`, `auth-bff`, `marketplace-api`, `otto` | `0` |
| `mark8ly-onboarding`, `admin`, `storefront` | live ports |

All four Go services require boot secrets CI does not have; only the Next.js apps, configured at build time, carry a live smoke port. The plan now says `0`, and explains why, so the next Go service added does not repeat it.

`smoke_path` stays `/healthz` rather than copying the siblings' `/health`: that is the endpoint this service genuinely serves, the path is unused while the port is `0`, and a truthful value is what someone enabling the check later actually wants.

Also corrected: the plan asserted `"target": "runtime"`; the Dockerfile's stage is `AS server`. The plan had told the implementer to verify that rather than trust it, which is the only reason it was caught — that instruction is now a statement of fact instead.

### DIFF
--- a/docs/superpowers/plans/2026-09-05-mcp-catalog-deployment.md
+++ b/docs/superpowers/plans/2026-09-05-mcp-catalog-deployment.md
@@ -376,6 +376,17 @@ The log must show the catalog seed POSTed. **A Job that completes is not the sam
 
 Query the registry for the connector. The API requires credentials, so this may need an operator; if you cannot query it, say so plainly rather than assuming the POST landed.
 
+- [ ] **Step 4: Confirm the route appears — this is the real acceptance**
+
+**No manual AgentGateway configuration is needed anywhere.** `agentgateway-route-sync-controller` (running the `agentic-registry` image in `agentgateway-system`) polls `REGISTRY_URL=.../v0/export/...` every **30 seconds** and reconciles routes into `TARGET_NAMESPACE=agentgateway-system` with `FIELD_MANAGER=agentgateway-registry-sync`. Its static ConfigMap carries only logging config; routes arrive by xDS. So publishing the seed and bumping the nonce genuinely IS the whole registration path.
+
+Wait ~30s after Step 2 and confirm the connector's route exists in `agentgateway-system`. If it does not appear within a couple of poll intervals, the seed did not register — go back to Step 2 rather than adding gateway config by hand, because hand-added config is exactly what the next reconcile will remove.
+
+**Two operational facts worth knowing before you touch this:**
+
+- **`PRUNE=true`.** The registry is authoritative: a route not present in the export is DELETED. Never hand-create a route to "help" — it will vanish at the next poll, and the disappearance will look like a different bug.
+- **`MIN_RESOURCES=27`** is a safety floor on the export. Adding this connector moves the estate's count up, not down, so it is not a concern here — but be aware the controller has a guard against acting on a short or failed fetch, and do not treat a quiet controller as a successful sync. Read its logs.
+
 ---
 
 ## Out of scope
--- a/docs/superpowers/plans/2026-09-05-mcp-catalog-deployment.md
+++ b/docs/superpowers/plans/2026-09-05-mcp-catalog-deployment.md
@@ -349,7 +349,32 @@ The tool list in the seed and the tool list the server returns must match. mark8
 git commit -m "feat(registry): register the mark8ly catalog connector"
 ```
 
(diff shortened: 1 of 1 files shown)


mahesh-sangawar:
Second commit pushed — following Kora further answered the routing question, and turned up one more gap.

**Registration is a registry publish, not gateway config.** `charts/apps/devai-registry-bootstrap` runs a Job that clones devai, reads `architecture/registry-seeds/`, and POSTs each YAML to the agentic registry's v0 API (`agentregistry.agentregistry-system.svc:12121`), upserting on `(kind, name, namespace)`. Discovery is registry-driven — "onboarding a new MCP is a publish, not a code change", as devai's hub documents.

**That exposed a gap in Task 5.** Adding the seed file to devai does nothing on its own: the bootstrap re-POSTs only when `reseedNonce` changes. New **Task 5b** bumps it in tesserix-k8s and verifies the Job's *log*, not its exit code — a Job that completes is not the same as a seed that registered.

**And it corrects a claim I made in the first draft.** I wrote that the seeds looked dormant because zero `MCPServer` CRs exist in the cluster. That inference was wrong — seeds are never meant to become CRs. The absence is expected, not evidence. The out-of-scope note has been narrowed accordingly: the `registry.solo.io` vs `kagent.dev` mismatch is real but concerns the third-party *catalog* seeds, not the product records this plan adds.

**One thing worth flagging about following Kora on deployment shape:** `kora-mcp`'s ArgoCD app points at `path: charts/apps/mcp-gateway` — the same shared chart, and therefore the same `slm-support-platform` image, as `mark8ly-mcp`. Kora has not solved the god-server problem; it is another tenant of it. So Kora's *auth architecture* is the thing to copy here, and its *deployment shape* is precisely what this work moves away from. The plan's own chart task stands.

PR #681, opened 2026-09-05
fix(ci): the connector image name was reversed, standard says mark8ly-catalog-mcp

mahesh-sangawar:
The connector's image name was reversed. `australis`'s product MCP onboarding guide names the server **`mark8ly-catalog-mcp`**; #677 published it as `mark8ly-mcp-catalog`.

Fixing it now, while nothing consumes it — an hour ago this became a real published package, and every later reference (chart, ArgoCD app, registry seed, gateway route) derives from this name.

## What the standard says

The guide's Mark8ly decision summary is explicit:

| Question | Direction |
| --- | --- |
| Server name | `mark8ly-catalog-mcp`, distinct from legacy `mark8ly-mcp`. |

and it names the downstream artifacts to match — `charts/apps/mark8ly-catalog-mcp/`, `architecture/registry-seeds/mcp-servers/mark8ly-catalog-mcp.yaml`.

## Deliberately narrow

**Only the image name changes.** Two things that look like they should follow it, and must not:

- **The GSM secret stays `prod-mark8ly-mcp-catalog-key`.** The guide pairs `MARK8LY_CATALOG_MCP_KEY` *from* `prod-mark8ly-mcp-catalog-key` — the secret genuinely keeps the other word order. A blind find-and-replace would have corrupted it, which is why this was done by hand.
- **No code moves.** The guide confirms every path already in `main` — `services/mcp/cmd/mcp-catalog/`, `internal/catalog/`, `internal/config/`, `internal/server/`, `go.mod`, `Dockerfile` — and states plainly: *"The merged Mark8ly layout is already in the right ownership boundary. Do not move it into `slm-support-platform` or `australis/servers/`."*

## Verified

`python3 .github/scripts/test-ci-contract.py` passes 6/6, with the closed image set and expected tuple both updated. The diff is three lines across two files.

## The old package

`mark8ly-mcp-catalog` has one published version (`main-de52104`) and nothing references it. It can be left to age out, or deleted by an operator — deleting is a package-registry action rather than a repo change, so it is not in this PR.

## Follow-up

The deployment plan (#675, corrected in #676) still needs a larger pass against the guide — it carries the old name, points ArgoCD at `argocd/prod/apps/mark8ly/` where the guide says `ai-apps/`, plans an eighth subscription on the shared Kargo warehouse where the guide requires an independent one, under-specifies mesh policy, and has nothing matching the guide's Gates A–G. That is a separate PR so this rename can land immediately.


mahesh-sangawar:
Closing unmerged — superseded rather than wrong.

This renamed the connector image to `mark8ly-catalog-mcp`, per the onboarding guide's Mark8ly decision summary. The packaging decision recorded in [australis ADR-0005](https://github.com/tesserix/australis/blob/main/docs/adr/0005-mark8ly-catalog-on-the-shared-gateway.md) means the catalog tools ship as tenant tools inside `slm-support-platform/services/mcp-gateway` (slm-support-platform#23) rather than as a product-owned service — so there is no connector image to name.

#687 removes the image entry entirely, which subsumes this.

Worth preserving from the analysis here, since it will bite whoever next touches these names: the GSM secret is `prod-mark8ly-mcp-catalog-key` while the gateway secret key is `MARK8LY_CATALOG_MCP_KEY` — the two genuinely use different word orders, and a find-and-replace across them corrupts the secret reference.

Issue #709, opened 2026-09-05
Rotate the Zitadel client secrets and login-client PAT exposed during the migration


mahesh-sangawar:
Two Zitadel client secrets were printed to a terminal during the migration work on 2026-09-04, and the `mark8ly-zitadel` login-client PAT was used extensively from a workstation throughout 2026-09-04/05.

## What that PAT did during the session

It is instance-level and can mint a session for any user of any product on the shared instance — the package doc calls it "the most powerful credential this service holds". In this session it was used to:

- create human users (`demo@mark8ly.com`, and attempts for others)
- create project grants on `mark8ly-admin`
- read the full TESSERIX org user list (16 users)
- probe the password complexity policy and error ids
- read IDP templates and organisation metadata

None of that is misuse — it is the work — but the exposure window is wide and the blast radius is the whole instance.

## Ask

1. Rotate the two client secrets that were displayed.
2. Rotate `prod-mark8ly-zitadel-login-client-token` (GCP Secret Manager). Consumed by `mark8ly-auth-bff` and, since tesserix-k8s#972, `mark8ly-platform-api` — both read it from the `mark8ly-zitadel` ExternalSecret, so a rotation propagates on the ESO refresh interval (1h) or a forced sync.
3. Consider whether day-to-day operational work should use a narrower credential than the login-client PAT. It could not delete a user during this session (`AUTH-AWfge`), so it is not omnipotent — but it can create users and grants, which is most of the risk.

## Note for whoever does it

Verify auth-bff and platform-api both pick up the new value before considering it done — a stale token fails closed at `ValidateZitadel`, so the failure mode is a pod that will not boot rather than silent degradation. That is recoverable, but it is a rollout, not a config tweak.

mahesh-sangawar:
## Done — all three credentials rotated and verified in production

Rotated 2026-09-05. Every claim below was verified against Zitadel/GCP/the cluster, not inferred.

### What was rotated

| Credential | Action | Old value now |
|---|---|---|
| `mark8ly-admin-web` client secret (app `389070493052829711`) | regenerated | `400 unauthorized_client` |
| `mark8ly-storefront-web` client secret (app `389070494764105743`) | regenerated | `400 unauthorized_client` |
| `mark8ly-login-client` PAT (`389070654164434959`) | new PAT `389408980633387523` issued, old one **deleted** | `401` |

GCP Secret Manager version 1 of all three secrets is **disabled**, so the exposed values cannot be read back.

### Rollout

[...]

Issue #686, opened 2026-09-05
Mobile admin app is GIP-only, so merchants who exist only in Zitadel cannot sign in on mobile


mahesh-sangawar:
The mobile admin app authenticates against GIP only. Merchants now exist in Zitadel, so mobile sign-in is broken for them.

## Evidence

`packages/mobile-shared/auth/gip.ts` is built on `@react-native-firebase/auth` (Firebase/GIP SDK). `apps/mobile-admin/app/login.tsx` consumes it via `@repo/mobile-shared/auth/provider`.

Grepping `packages/mobile-shared` for `zitadel` returns **nothing** — there is no alternative path and no provider flag. Unlike the web apps, mobile has no branch to flip.

For comparison, after the cutover:

| Surface | Provider-aware |
|---|---|
| `apps/admin` | yes (9 files branch on the flag) |
| `apps/storefront` | yes (5 files) |
| `apps/onboarding` | **no** — see #685 |
| `apps/mobile-admin` | **no** — this issue |

## Why it is harder than the web fixes

#679 and #685 are branches inside a Next.js app: flip a flag, redeploy. This is a native app:

- Firebase's SDK is not a drop-in for Zitadel — it needs an OIDC/PKCE flow, most likely through auth-bff, not a client SDK swap
- shipping means an app store release, so the fix cannot be rolled out on the usual cadence, and old installs keep pointing at GIP until users update
- phase 4 of #524 already dealt with mobile *tenancy* (`X-Acting-Tenant-Id`, validated against FGA), so the tenancy half is done — this is only the identity half

## Severity

Depends on real mobile usage, which I have not measured — worth checking before planning. Two things are true regardless:

- a merchant whose account exists only in Zitadel cannot sign in on mobile at all
- because old installs cannot be forced to update, GIP may need to stay alive for mobile after it is retired everywhere else, which affects when GIP can actually be switched off

That second point is the one to weigh now: it is a constraint on the #524 endgame, not just a bug.

## Not yet verified

I have not run the mobile app against the current backend. This is inferred from the code — no Zitadel path exists in `mobile-shared`, so there is nothing for it to call. Worth a real sign-in attempt on a device before planning the work.

mahesh-sangawar:
## Scope correction — this is smaller than the issue implies

[...]

PR #728, opened 2026-09-05
feat(auth-bff): add the OIDC authorization-code token exchange (#686)

mahesh-sangawar:
Part of #686. First step of keeping mark8ly's own login form on mobile rather than sending merchants to Zitadel's hosted login.

## Why this is new code rather than reuse

There is **no authorization-code exchange anywhere in the estate**. `apps/admin/app/auth/callback/route.ts` says so outright:

> It does NOT exchange `code` for anything — there is nothing left to exchange it for, and it must not try.

The web admin rides auth-bff's own `m8_session` cookie; the OIDC flow is used only to drive Zitadel's session APIs and produce a callback URL. Nothing ever holds a Zitadel access token.

Mobile cannot work that way: marketplace-api's mobile admin routes verify a **bearer JWT** against Zitadel's JWKS, and a session cookie is not a bearer token. So keeping our own form requires turning the code into a token server-side. That is what this adds.

## Two things that would each have cost hours

**The project-audience scope.** `ZitadelVerifier` pins `aud` to the mark8ly-admin project so a storefront token — same issuer, same signing key, same human — cannot be replayed as an admin credential. That audience only exists if the **authorize** step requested `urn:zitadel:iam:org:project:id:<id>:aud`. Omit it and everything works right up until the API rejects a perfectly valid token with a flat 401 that reads exactly like a wrong password. `TestAuthorizeURL_RequestsTheProjectAudienceScope` pins it.

**`offline_access`.** Without it there is no refresh token, and a merchant is silently signed out when the access token expires (~1h). Also pinned.

## Notes

- **`client_secret_basic`**, matching the app's `authMethodType: BASIC`. The secret is deliberately not duplicated into the form body, so it cannot land in request-body logging — there is a test asserting its absence.
- `redirect_uri` must byte-match the auth request's or Zitadel returns a generic `invalid_grant` that names nothing; the doc comment says so at the call site.
- An `error` on the callback is returned as an error rather than an empty code, so a caller cannot treat it as success and fail confusingly one layer later.
- The upstream failure body is preserved: `invalid_grant` / `invalid_client` / `unauthorized_client` point at three completely different misconfigurations.

## Credentials

Uses the mark8ly-admin confidential client's id and secret, which the chart **already injects** into auth-bff as `ZITADEL_ADMIN_CLIENT_ID` / `ZITADEL_ADMIN_CLIENT_SECRET`. Before this, no code read them (see the correction on #709) — so this is the first consumer, and no chart change is needed.

## Scope

This PR adds the capability only; nothing calls it yet, so it is inert on deploy. `gofmt`/`vet`/`build` clean, full `go test ./...` green, 4 new tests.

**Still to come for mobile login:** a public login endpoint (recommend proxying via marketplace-api so auth-bff keeps its `X-Internal-Auth` protection), an **email-OTP screen** — a fresh install is always a new device, so `email_otp_required` will fire on essentially every first sign-in — and the MFA/TOTP path.

### DIFF
--- a/services/auth-bff/internal/zitadellogin/auth_request_test.go
+++ b/services/auth-bff/internal/zitadellogin/auth_request_test.go
@@ -1,0 +1,100 @@
+package zitadellogin
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+// CreateAuthRequest is what lets the server start an OIDC flow with no
+// browser involved — the precondition for keeping mark8ly's own login
+// form on mobile instead of redirecting to Zitadel's hosted login.
+func TestCreateAuthRequest_ExtractsIDFromTheRedirect(t *testing.T) {
+	var gotQuery string
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		gotQuery = r.URL.RawQuery
+		// Zitadel answers /oauth/v2/authorize with a 302 to the login UI,
+		// carrying the auth request id. Kept deliberately short and
+		// obviously fake: a realistic id is a long digit string that
+		// secret scanners classify as a high-entropy credential.
+		w.Header().Set("Location", "https://admin.mark8ly.com/login?authRequest=V2_test_auth_request")
+		w.WriteHeader(http.StatusFound)
+	}))
+	defer srv.Close()
+
+	ex := NewTokenExchanger(srv.URL, "client-1", "s", srv.Client())
+	id, err := ex.CreateAuthRequest(context.Background(),
+		"https://admin.mark8ly.com/auth/callback", "389070376568619523", "state-1")
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if id != "V2_test_auth_request" {
+		t.Fatalf("id = %q", id)
+	}
+	if !strings.Contains(gotQuery, "aud") {
+		t.Fatalf("the authorize call must carry the project-audience scope, got %q", gotQuery)
+	}
+}
+
+// The redirect must NOT be followed: following it would fetch the login
+// page and lose the id, and on a real deployment would leave the server
+// chasing a browser-facing URL.
+func TestCreateAuthRequest_DoesNotFollowTheRedirect(t *testing.T) {
+	var hits int
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		hits++
+		if r.URL.Path == "/oauth/v2/authorize" {
+			w.Header().Set("Location", "/login?authRequest=V2_1")
+			w.WriteHeader(http.StatusFound)
+			return
+		}
+		w.WriteHeader(http.StatusOK)
+	}))
+	defer srv.Close()
+
+	_, err := NewTokenExchanger(srv.URL, testClientID, testClientPlaceholder, srv.Client()).
+		CreateAuthRequest(context.Background(), "https://x/cb", "p", "st")
+	if err != nil {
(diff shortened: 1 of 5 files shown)


mahesh-sangawar:
Pushed a second commit: `CreateAuthRequest` — starting an OIDC flow server-side with no browser.

This is the other precondition for keeping our own form. `/auth/zitadel/login` requires an `auth_request_id`, and on web the browser gets one by being redirected through `/oauth/v2/authorize`. A native app posting credentials to an API has no such redirect, so the server has to create the request itself.

Zitadel answers `/oauth/v2/authorize` with a 302 to the login UI carrying `?authRequest=<id>`. The redirect is deliberately **not** followed — following it fetches a browser-facing login page and loses the id. There is a test asserting exactly one request is made, because "it works" and "it works for the right reason" differ here: a client that follows redirects would still return *something*.

The no-redirect policy is applied to a **copy** of the http.Client, so it cannot leak into a shared client other callers depend on.

Together with the exchange in the first commit, the server can now drive a full flow headlessly: create auth request → password session → finalize → `callback_url` → code → tokens.

`Result.CallbackURL` was already returned to internal callers on a clean login (`{"callback_url": ...}`), so no change to the existing login path was needed.

4 more tests, including a 302 with no id and a non-redirect response — the latter preserves the upstream text, because a `redirect_uri` mismatch is the likeliest misconfiguration and is the only thing that says which value was wrong.

mahesh-sangawar:
Third commit: the mobile login and TOTP routes. Also rebased onto current main (the earlier diff looked noisy only because main had moved).

## The design decision worth reviewing

The mobile routes differ from the web ones in **exactly one respect**: a completed login answers with tokens instead of `callback_url`. Everything before that point is the *same code path*, deliberately:

- the wrong-user / wrong-password collapse that stops account enumeration
- re-resolving the subject's email from Zitadel via `SessionFactors` + `UserEmail` rather than trusting the request body — the property that stops a caller with valid credentials of their own from having a sign-in code mailed to a victim
- the post-identity gauntlet (FGA membership, device recognition)
- the MFA / email-OTP step-up gate

Those are security properties, and reimplementing them per surface is how they drift apart. So `login`/`totp` became thin wrappers over `loginMode`/`totpMode` with an `issueTokens` flag, and only the final branch of `finishComplete` differs.

**Separate routes, not a request flag.** `issue_tokens` in the body would have been less code, but it would let any internal caller opt itself into token issuance. Distinct routes make the surface explicit.

## Fail-closed choices

- **No issuer configured → 500**, never a fallback to `callback_url`. A mobile client cannot use a callback URL, so the fallback would look like a successful login right up until the first API call 401s — the hardest possible failure to diagnose from a device.
- **A step-up outstanding → no tokens.** Handing tokens over while an OTP or TOTP is pending would defeat the gate entirely. Both cases are tested, and the OTP one is the *common* path, not an edge case: a fresh install is always an unrecognised device.
- **`callback_url` is never returned to mobile** — the client has no use for it and it carries a live authorization code.

Google/IDP sign-in passes `issueTokens=false` with a comment saying why: the app's Google button still goes through the Firebase SDK, so there is no mobile surface for it yet.

## Verification

6 new tests here (11 across the PR). The one I'd point a reviewer at is `TestWebLoginStillReturnsCallbackURLWhenAnIssuerIsConfigured` — it pins that configuring the issuer does not change the web contract, which is the actual regression risk in this change. The full auth-bff suite is green, including the arch tests that pin `finalize` to a single call site behind the `sufficient` witness.

Still inert on deploy: nothing constructs a `TokenExchanger` yet, so `WithTokenIssuer` is never called and the mobile routes answer 500 if reached.

## Next

The public front door on marketplace-api (resolving `workspace_tenant` by email via `ListMyTenants` before proxying, so the client does not need to know it), then the app's OTP screen.

PR #740, opened 2026-09-05
feat(auth-bff): wire the mobile token issuer from config (#686)

mahesh-sangawar:
Part of #686. Wires the mobile token issuer, which #728 deliberately left inert.

## Found by testing production, not by reading code

With #728 deployed I ran the real end-to-end call:

```
POST https://api.mark8ly.com/api/v1/mobile/admin/auth/login   → 502
```

Chasing it through the logs showed the whole chain working and stopping at exactly one place:

- `api.mark8ly.com` → marketplace-api ✅
- **tenant resolved by email** — it got past the no-tenant 404, so `ListMyTenants` found Mumbai Spice Co ✅
- marketplace-api → auth-bff with valid internal auth (a 500, not a 401) ✅
- auth-bff's mobile route reached ✅
- `zitadellogin: could not create an auth request for mobile login err="no token issuer configured"` ⛔

That last line is #728's documented "inert on deploy" state, not a bug. This PR supplies the wiring.

## Config

`ZITADEL_ADMIN_CLIENT_ID` and `ZITADEL_ADMIN_CLIENT_SECRET` are the mark8ly-admin confidential client's, and **the chart already injects both** — until now nothing read them (see the correction on #709). `ZITADEL_ADMIN_PROJECT_ID` is already set too.

`ZITADEL_ADMIN_REDIRECT_URI` is the one genuinely new value, and it is deliberately **not defaulted**. It must byte-match the redirect the auth request was created with or Zitadel refuses the exchange with a generic `invalid_grant` that names nothing — and a hardcoded default would be silently wrong in any environment that is not production.

## Deploy ordering

Empty config leaves mobile token issuance **off**, which is exactly today's behaviour. So this lands code-first safely, and a chart change turns it on — the right way round, since it is *adding* required config that would otherwise fail pods at boot.

A companion tesserix-k8s PR adds `ZITADEL_ADMIN_REDIRECT_URI: https://admin.mark8ly.com/auth/callback` to the auth-bff chart. **Merge this first**, then the chart.

## All four or nothing

Half-configured means OFF, not half-on. A login that "succeeds" without a usable token is indistinguishable from a working one until the first API call 401s — from a device, the hardest possible failure to diagnose. Refusing at 500 is loud and immediate. Five tests pin that, one per missing value, plus the unconfigured default.

## Verification

`gofmt`/`vet`/`build` clean, full auth-bff suite green. After this deploys I will re-run the production call; the expected answer for a never-seen device is `email_otp_required`, not tokens.

### DIFF
--- a/services/auth-bff/cmd/server/main.go
+++ b/services/auth-bff/cmd/server/main.go
@@ -102,7 +102,20 @@ func newZitadelHandlers(cfg *config.Config, zitadelClient *zitadellogin.Client, 
 	if err != nil {
 		return nil, err
 	}
+	// Mobile token issuance (#686). Enabled only when all four values are
+	// present; any missing one leaves the mobile routes refusing with 500
+	// rather than half-configured, because a login that "succeeds" without
+	// a usable token is indistinguishable from a working one until the
+	// first API call 401s.
+	var mobileTokens *zitadellogin.TokenExchanger
+	if cfg.ZitadelAdminClientID != "" && cfg.ZitadelAdminClientSecret != "" &&
+		cfg.ZitadelAdminRedirectURI != "" && cfg.ZitadelAdminProjectID != "" {
+		mobileTokens = zitadellogin.NewTokenExchanger(
+			cfg.ZitadelIssuer, cfg.ZitadelAdminClientID, cfg.ZitadelAdminClientSecret, nil)
+	}
+
 	merchant := zitadellogin.NewHandler(zitadelClient, complete).
+		WithTokenIssuer(mobileTokens, cfg.ZitadelAdminRedirectURI, cfg.ZitadelAdminProjectID).
 		WithHostedLoginBaseURL(cfg.ZitadelIssuer).
 		WithInternalAuth(cfg.MarketplaceInternalAuthSecret).
 		WithReturnURLAllowlist(adminReturnURLs).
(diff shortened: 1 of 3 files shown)


mahesh-sangawar:
## Correction to this PR's description

It says `ZITADEL_ADMIN_CLIENT_ID` / `_SECRET` are "already injected by the chart". **That was true when I wrote it and is no longer true.**

`tesserix-k8s#1007` — *"drop four zitadel client env vars no code reads"*, merged today at 20:55 — removed them. That change was correct on its own terms, and it was acting on the finding I posted to #709 establishing that nothing read them. It simply landed while I was writing the code that does.

Verified against the live deployment: `mark8ly-auth-bff` currently has only `ZITADEL_LOGIN_CLIENT_TOKEN` among the Zitadel secret refs.

**So this PR needs all three values supplied, not one.** tesserix-k8s#1015 restores the id/secret pair (with a comment naming `ZitadelAdminClientID` as the reader, so they are not dropped a third time) and adds `ZITADEL_ADMIN_REDIRECT_URI`.

Nothing about this PR's code changes — the "all four or nothing" guard already handles missing values by leaving the feature off, which is exactly what has been happening in production since #728 deployed. Merge order is unchanged: this PR first, then the chart.

Worth noting the shape of the near-miss: had the guard been "enable if *some* values are present", this would have half-configured itself and produced logins that succeed with unusable tokens.

Issue #792, opened 2026-09-07
Migrate (or delete) apps/mobile-storefront and apps/storefront-mobile — both are 100% GIP


mahesh-sangawar:
Split out of #708. These two apps were **never migrated** and the issue's inventory does not mention them.

Verified 2026-09-07 on `main`:

```
apps/mobile-storefront : zitadel refs = 0
apps/storefront-mobile : zitadel refs = 0
apps/mobile-admin      : zitadel refs = 21
```

Both are 100% GIP/Firebase, with `gipTenantId` a **required build field**. GIP cannot be switched off while they are the only auth these apps have.

## Open questions to settle first

- **Are both apps actually alive?** Two similarly-named storefront mobile apps is itself suspicious — one may be superseded. If one is dead, delete it rather than migrating it; that is much cheaper.
- **Has either ever shipped?** The admin app's answer was instructive: no iOS release was ever published, and Play shows `0+` downloads. If these have never shipped either, there is no installed base to protect and the cutover is free.

## Scope, once those are answered

Mirror what `apps/mobile-admin` already does: the Zitadel sign-in client, session storage, the auth gate, and the api-client reading the Zitadel session. That work is done and proven for admin — this is porting it, not designing it.

Depends on #787 (the storefront customer verifier on marketplace-api) — these apps authenticate customers, and there is no Zitadel customer path server-side yet.

## Done when

- [ ] Confirmed which of the two apps is live.
- [ ] Live app(s) sign in through Zitadel, verified on a device.
- [ ] `gipTenantId` no longer required to build them.
- [ ] Any dead app deleted rather than migrated.

See `docs/auth/2026-09-07-gip-removal-audit.md`.

mahesh-sangawar:
## Neither app has ever shipped — so this blocks nothing, and one of the two should be deleted rather than migrated

Investigated 2026-09-07.

### The decisive fact: no installed base for either

```
App Store   com.mark8ly.store      -> 0 results
App Store   com.mark8ly.storefront -> 0 results
Play Store  com.mark8ly.store      -> HTTP 404
Play Store  com.mark8ly.storefront -> HTTP 404
Play Store  com.mark8ly.admin      -> HTTP 200   (control: a real listing)
```

Neither storefront app exists on either store. There is **nobody holding a GIP-only build of either**, so they do not constrain the GIP switch-off at all. This issue is not a blocker on #708 — it is cleanup that can happen whenever.

### They are not two versions of one app; they are two different bets

[...]

PR #802, opened 2026-09-07
chore(storefront-mobile): delete the superseded single-tenant storefront app (#792)

mahesh-sangawar:
Part of #792, per the owner's decision: **keep `apps/mobile-storefront`** (multi-tenant white-label), **delete `apps/storefront-mobile`** (single-tenant). 88 files removed.

## Why this one

| | `storefront-mobile` (deleted) | `mobile-storefront` (kept) |
|---|---|---|
| Commits | 45 — but **only renovate bumps since 2026-04-10** | 1, from the white-label rearchitecture |
| EAS project | `"your-eas-project-id"` — **placeholder, never provisioned** | per-merchant, by design |
| Identity | static `com.mark8ly.store` | derived per merchant from `branding/<slug>/config.json` |
| Shipped? | **no** | **no** |

Neither app ever shipped:

```
App Store   com.mark8ly.store / com.mark8ly.storefront  ->  0 results
Play Store  com.mark8ly.store / com.mark8ly.storefront  ->  HTTP 404
Play Store  com.mark8ly.admin                           ->  HTTP 200   (control)
```

So there is no installed base on either side, and this deletion strands nobody. `storefront-mobile` had four months of dependency churn on an app that was never built through EAS and never released — pure maintenance cost. The white-label direction is corroborated by the Pro App lifecycle billing in `marketplace-api` (`internal/whitelabel/*`), which only makes sense if merchants get their own apps.

## What is NOT done here

`mobile-storefront` is still 100% GIP and its sign-in currently **throws** — #799 removed the shared Firebase backend it depended on. That is tracked in #792, which stays open for the migration half. This PR only removes the superseded app.

## About the lockfile — please let CI be the judge

Deleting a workspace changes `package-lock.json`. I regenerated it with `npm install --package-lock-only --ignore-scripts`, which removed all 56 `apps/storefront-mobile` references (diff: −1471/+188, mostly the deleted subtree plus some `@typescript-eslint` churn).

**I could not verify `npm ci` locally** — it 401s on `@tesserix/otto-widget` from GitHub Packages without `NODE_AUTH_TOKEN`, which this environment does not have. CI runs `npm ci`, so that is the real gate. If it fails on the lockfile, that is why, and the fix is a regeneration in an environment holding the token — **do not merge this on a red lockfile check.**

Nothing else referenced the app: grepped `*.json`, `*.yml`, `*.yaml`, `*.ts`, `*.tsx`, `*.js` — only `package.json`/`package-lock.json` (the `apps/*` workspace glob) and its own files. It has no CI workflow of its own.


mahesh-sangawar:
## Correction to the PR body, and the CI fix

The PR body says *"Nothing else referenced the app"*. **That was wrong**, and it broke CI:

```
x No package found with name '@repo/storefront-mobile' in workspace
```

Root `package.json` had four turbo scripts (`build`, `lint`, `test`, `check-types`) each carrying `--filter='!@repo/storefront-mobile'`, and turbo errors when a filter names a package that no longer exists.

**How I missed it:** I grepped for the directory name `storefront-mobile`, and `package.json` *did* appear in those results — I assumed it was the `apps/*` workspace glob and never read the matching line. The reference was by **package name** (`@repo/storefront-mobile`) in a turbo filter, which the same grep would have shown me had I looked at the line rather than the filename.

Fixed in `f4d99a78`:
- removed the four `--filter='!@repo/storefront-mobile'` flags (the `@repo/mobile-storefront` filters are **kept** — that app survives);
- removed the README tree entry and repaired the `└──` connector on the line that is now last.

Verified locally by reproducing the exact failing command:

```
npx turbo run check-types --filter='!@mark8ly/platform-api' ... --filter='!@repo/mobile-admin' --dry=json
EXIT=0
```

The lockfile check — the risk I *did* flag — was not the failure. It passed.

PR #835, opened 2026-09-07
refactor: retire per-tenant SSO and the pricing claim that advertised it (#820)

### LINKED ISSUE #820: Per-tenant SSO is unwired on both the config and login sides, so break-glass guards a door that does not open yet
Found while mounting the break-glass login route (#642, PR #819). Break-glass exists to rescue a tenant whose own SSO has broken — so it is worth stating plainly that **per-tenant SSO does not appear to be wired at all**, on either the config side or the login side.

## What exists

- Migrations: `000070_tenant_sso_configs`, `000071_tenant_sso_user_mappings`
- `internal/handlers/admin/SSOConfigHandler` — the admin-side config CRUD
- `internal/handlers/public/sso_login.go` — login / callback / logout
- `plangate.FeatureSSO`, enabled on exactly one plan tier and `Disabled` on the other three
- `authbffclient.HTTPIssuer`, which #642 built and which `sso_login.go` needs

## What is missing

**Login side** (`internal/handlers/public/sso_login.go`):
- `TenantResolver` has no production implementation anywhere. Its own doc comment is a literal `TODO(wiring)` (`:37`).
- `OIDCRPs`, the per-tenant relying-party map, is never constructed at startup.
- `loginSAML` is a no-op stub with a second `TODO(wiring)` (`:218`).

**Config side** (`internal/handlers/admin/routes.go:160-173`):
- The routes mount only when `deps.SSOConfigHandler != nil`, and **`NewSSOConfigHandler` is never called in `main.go`**. So a merchant on the tier that pays for SSO has no way to configure it.

This is the same shape as the break-glass login bug #642 was opened for: a handler that exists, is tested, is registered behind a nil-check, and is never constructed — so the route silently does not exist. `#819` fixed that instance; these two are the same pattern still open.

## Why it matters beyond the missing feature

#642's premise is that a Pro/Enterprise tenant attaches their own IdP, that IdP breaks, and their admins are locked out — so break-glass gives them one emergency local account. If no tenant can attach an IdP in the first place, break-glass currently protects against a failure mode that cannot occur yet.

That does not make #642 wasted — the machinery is built, tested and now mounted, and it will be needed the moment SSO ships. But it does mean **an operator should not be told mark8ly supports tenant SSO today**, and it probably reorders priorities: the escape hatch is finished while the door it guards is not.

## Suggested scope

1. Decide whether per-tenant SSO is still wanted. If not, the honest move is to remove the surface (handlers, plan feature, migrations) rather than leave a half-built feature that reads as shipped — the same reasoning that deleted the retired GCP Secret Manager client in #813.
2. If it is wanted: construct `SSOConfigHandler` in `main.go` so the config surface exists, implement `TenantResolver` against `internal/tenants` or tenant-service, build `OIDCRPs` at startup, and either implement or explicitly remove `loginSAML`.
3. Add a mount test for both surfaces, driven through the real router. Note the 405-vs-404 probe does **not** work in this estate: `HandleMethodNotAllowed` is never set in any service, so gin answers 404 whether or not a route is mounted, and that probe distinguishes nothing.

## Verification notes

Established by reading, not inference: `grep` for `NewSSOConfigHandler` returns no hit in `cmd/marketplace-api/main.go`; `TenantResolver` has no implementing type; `OIDCRPs` has no assignment outside tests.

mahesh-sangawar:
Closes #820. Two commits: the code, then the customer-facing claim. **Both halves are required** — removing the implementation while still advertising the feature would be worse than either alone.

## Why retire rather than finish

Per-tenant SSO was unreachable from both directions and could not have been made to work without a rewrite:

- **`NewSSOConfigHandler` and `NewSSOLoginHandler` were never called from `main.go`.** The routes mount behind nil-checks, so a merchant on the tier that *paid* for SSO had no way to configure it and no way to use it.
- `TenantResolver` had **no implementation** — its own doc comment was a literal `TODO(wiring)`. `OIDCRPs` was never constructed. `loginSAML` was a no-op stub.
- **The GIP provisioning client was already deleted** in #796 ("delete the unreachable GIP-backed SSO provisioning client"), so the layer that creates IdP configs was gone before this PR. What remained could not provision a provider at all.
- Both tables held **0 rows**, and there is no merchant-facing configuration UI anywhere.

Finishing it meant rebuilding provider management against Zitadel's IDP APIs and building a config UI from scratch — a project, not a wiring task.

## The customer-facing half

SSO was advertised in three places, so the claim goes with the code:

- `apps/onboarding/components/marketing/Pricing.tsx` — the **public pricing page**: `'SSO (SAML / OIDC)'` removed from the Pro feature list
- `apps/admin/lib/copy/pricing.ts` — same line, in-app
- `apps/onboarding/app/help/page.tsx` — *"Pro adds up to 10 stores, a full read/write API, ~~SSO,~~ and priority support."*

This is free right now and will not stay free: `store_subscriptions` has **0 rows** and Stripe is still in test mode, so **no customer has bought Pro expecting SSO**. That window closes the moment billing goes live (#366/#371).

## The coupling that had to be handled

**Break-glass's login route was gated on `plangate.FeatureSSO`** (`routes.go:161`, wired in #819). Removing the feature would have stranded that gate.

`RequireBreakGlassFeature` now takes a minimum **plan** instead of a feature and checks `PlanAtLeast(plan, subscription.PlanPro)` — behaviour-neutral, since `FeatureSSO` was enabled on `PlanPro` and nowhere else. The 403 shape, and the deliberate body-peek-and-restore (the route runs before auth middleware, so it reads `tenant_id` from the JSON body), are both preserved. A comment records why it changed.

All 7 break-glass tests pass, including `TestBreakGlassLogin_RouteIsMounted` and `TestRequireBreakGlassFeature_RefusesTenantWithoutFeature`.

**Break-glass itself is untouched.** Its premise weakens without SSO — it exists to rescue tenants whose own IdP breaks — but that is a separate decision, and #642/#404 should be revisited deliberately rather than swept along here.

## Fallout handled rather than worked around

- A new forward migration drops both tables; `000070`/`000071` are untouched, because deleting applied migrations breaks replay. `ExpectedSchemaVersion` moved with it.
- `plangate.RequireFeatureByTenant` had zero callers once `FeatureSSO` went — deleted rather than left as dead code.
- Two `tenantpurge` steps and a `customererasure` declared-exclusion referenced the dropped tables; the latter would have failed `TestDeclaredExclusions_AreJustifiedAndLive`.
- A test in `platformadmin` literally constructed `admin.NewSSOConfigHandler(...)`, which would have broken `go build`.
- Fixed a pre-existing latent bug in `matrix_test.go`: the assertion said `26` while its own failure message said "expected 25".
- Doc comments that pointed at deleted files or described "five admin route groups" corrected — a comment that misdescribes the code is how the next mistake gets made.

## Verified

`go build`, `go vet`, `go vet -tags=integration`, `go test ./...`, `gofmt -l` all clean. `npm ci` at the repo root, then **`npm run build` passed for both `apps/onboarding` (33 routes) and `apps/admin`** — that build is the only gate here that catches certain Next.js errors; vitest and tsc are not sufficient.

Every remaining `sso` hit is a comment: auth-bff's two out-of-scope mentions, break-glass's own premise docs, and comments explaining this retirement.


mahesh-sangawar:
**Heads-up: this now conflicts with #836, which merged seven minutes after this PR was opened and does the opposite.**

Not a criticism of the analysis here — every premise in the description was true when it was written. But #833 and #836 have since fixed each one, so the reasoning no longer holds:

| this PR's premise | state on main now |
|---|---|
| `NewSSOConfigHandler` / `NewSSOLoginHandler` never called from `main.go` | both constructed in `buildSSO` and mounted on both engines (#836) |
| `TenantResolver` has no implementation, its doc is a literal `TODO(wiring)` | `StoreSlugTenantResolver`, resolving through the stores projection (#833) |
| `OIDCRPs` never constructed | replaced by `sso.RelyingPartyCache`, built per tenant on first use (#833) |
| `loginSAML` is a no-op stub | answers 501, agreeing with the callback (#833) |
| the layer that provisions an identity is gone | `internal/ssousers` on both sides, resolving through Zitadel (#836) |

`mergeStateStatus` is already `DIRTY` / `CONFLICTING` for that reason.

## One premise that is still true

**There is no merchant-facing configuration UI.** #836 mounted the config API and the login flow; it did not build a screen. So SSO today is reachable by API and not by a merchant clicking anything — which is a real gap, and a fair reason to argue the feature is not shippable yet.

That is a smaller claim than this PR makes, though, and it argues for building a UI rather than deleting the engine.

## What needs deciding

The direction was settled deliberately while this was open — wire it rather than retire it — so I have not touched this PR beyond leaving this comment. Whoever owns it should decide between:

1. **Close this.** The code it removes is now wired, tested and mounted, and its premises are addressed.
2. **Keep the customer-facing half.** The second commit removes `'SSO (SAML / OIDC)'` from the public pricing page and the in-app copy. That claim is *still* overstated: the API works but no merchant can configure SSO without one, and there are no SSO tenants. Splitting that commit out and landing it alone would be honest either way.
3. **Retire after all**, in which case #836 should be reverted first rather than this rebased over it — a 3,000-line deletion resolved through conflicts against freshly-added code is the wrong way to remove a feature.

Option 2 is worth doing regardless of which way 1-vs-3 goes.

Refs #820, #833, #836.

mahesh-sangawar:
Closing without merging. **This is wrong now** — per-tenant SSO was implemented and mounted in #833 and #836 while this branch was being written, and merging it would delete working, tested, mounted code plus two bug fixes.

## Every gap this PR was built on is closed

| Claim this PR rested on | State on `main` |
|---|---|
| `TenantResolver` has no implementation (`TODO(wiring)`) | Real — `StoreSlugTenantResolver`, resolving via the stores projection |
| `OIDCRPs` never constructed | Real — `sso.RelyingPartyCache`, built lazily per tenant from OpenBao so one unreachable IdP cannot delay startup or affect another tenant |
| Neither handler constructed in `main.go` | Both built in `cmd/marketplace-api/wiring_sso.go`, both route sets mounted |
| `sso.UsersRepo` is a `TODO(wiring)` | Real — `ssousers.NewClient` against platform-api, feeding `sso.NewJITProvisioner` |

#836's wiring even answers this issue by name: *"Both surfaces are constructed here or neither is: a merchant able to SAVE an IdP config that nobody can then sign in with is the 'looks delivered, does nothing' state #820 was opened about."*

It also fixed a bug I had not spotted: `loginSAML` previously returned **200 with an empty body** — a blank page to a user, and a *successful login* to a monitor. It now returns 501, so the two SAML branches agree and neither can be mistaken for working.

## Where my evidence was stale

Worth recording, since it is the reason this happened. My argument was "GIP-based, therefore finishing it is a rewrite". `internal/sso/gip_client.go` had **already been deleted in #796**, which I only discovered during execution — so the GIP dependency I built the case on was largely gone before I started. The remaining objections (unwired, no resolver, no RP cache) were real, and have now simply been fixed rather than deleted.

## What does NOT get closed with this PR

Two findings from this work survive the implementation and are follow-ups, not retirement:

1. **The pricing claim is half false.** `apps/onboarding/components/marketing/Pricing.tsx` and `apps/admin/lib/copy/pricing.ts` advertise **"SSO (SAML / OIDC)"**, and SAML answers `saml_not_implemented` (501). Either implement SAML or narrow the copy to OIDC.
2. **There is no merchant-facing SSO configuration UI.** The config API is mounted, but nothing in `apps/admin` renders it, so a Pro merchant cannot self-serve — an operator has to call the API on their behalf.

Both are recorded on #820.

## Also not merged, and worth keeping

This branch contained a few incidental fixes that are unrelated to SSO and still valid: a latent bug in `plangate/matrix_test.go` (the assertion checked `26` while its own failure message said "expected 25"), and several doc comments pointing at files or route-group counts that had already changed. Small enough to redo if wanted; noting them so they are not silently lost with the branch.

Issue #857, opened 2026-09-08
CI runs every Node gate on 22 while production containers run 24 — and it is already hiding two failing tests


mahesh-sangawar:
Split out of #834, which is otherwise delivered. Found while confirming that issue's last open item — the reported "appCredentials.test.ts fails locally but CI is green" disagreement. It reproduces, and the cause is not the test.

## The divergence

| where | Node major | source |
|---|---|---|
| CI unit tests + `next build` | **22** | `.github/workflows/ci.yml:76` — `node_version: "22"` |
| CI e2e (runnable subset) | **22** | `.github/workflows/e2e-runnable.yml` — pinned to match ci.yml, deliberately |
| **Production containers** | **24** | `apps/admin/Dockerfile:11,32,106` — `base-node-builder-24`, `base-node-runtime-24` |

**Every gate runs on a Node major that production does not use.** CI is not testing the runtime that ships.

## What it is already hiding

`apps/admin/tests/unit/api/subscription/appCredentials.test.ts` — 2 of 8 tests fail on Node v24.20.0 and pass on 22:

```
FAIL  uploadAppleCredentials > sends multipart/form-data with issuer_id and key_id fields
FAIL  uploadGoogleCredentials > sends service_account_json as a multipart file field

ApiError: The expression evaluated to a falsy value:
  assert(typeof value === "string" && webidl.is.USVString(value) || webidl.is.File(value))
```

The assert is undici's, raised inside MSW's `request.formData()` in the mock handler — Node 24 tightened the webidl check on FormData values. The `204 — resolves void` test in the same file passes, because it never parses the body.

So this specific one looks like an MSW/undici test-infrastructure incompatibility rather than a product defect: `postMultipart` itself is likely fine, and the throw surfaces at `lib/api/subscription/appCredentials.ts:50` only because the handler died and the response was not a 204. **That should be confirmed rather than assumed** — it is exactly the kind of thing this gap is bad at telling us.

## Why it matters beyond this one test

The failure mode is the wrong way round. A Node-24-only *product* bug is invisible to every gate and ships; a Node-24-only *test* bug is invisible to CI and burns whoever runs the suite locally on a supported version. `CLAUDE.md` already states Node 24 for all Next.js apps, so 22 in CI is drift, not policy.

Same family as #834 itself: a check that exists, reads as enforcement, and is not exercising the thing it appears to guard.

## What this asks for

[...]

PR #859, opened 2026-09-08
test(admin): give the jsdom environment Web Storage, which Node 26 shadows

mahesh-sangawar:
Makes `MobileAppPrompt.test.tsx` pass on Node 26. One file, 54 added lines.

**Narrowed.** This PR previously also carried the CI Node bump and the `appCredentials` fix. #863 does both, with a more thorough diagnosis of the `File`-realm mismatch than mine, so those are dropped here rather than duplicated. What remains does not overlap #863 at all.

## The bug

```
v22.19.0  localStorage absent              MobileAppPrompt passes
v24.20.0  localStorage absent              passes
v26.5.0   present, undefined               8 tests fail
```

Node 26 defines `globalThis.localStorage` itself, as an accessor returning `undefined` unless the process was started with `--localstorage-file` — it warns exactly that. vitest's jsdom environment populates globals only for keys **not already present**, and the key IS present, so jsdom's real Storage never lands on it.

The tell is `hasOwnProperty("localStorage") === true` with a value of `undefined`: a key that exists and reads as nothing. Constructing a `JSDOM` directly gives `typeof window.localStorage === "object"`, so neither jsdom 30 nor this repo's config is at fault.

`MobileAppPrompt` persists its dismissal (`MobileAppPrompt.tsx:50,62`), and the test's `afterEach` calls `localStorage.clear()` — which throws, and takes `cleanup()` with it, so later tests inherit a mounted DOM and fail with *"Found multiple elements"*. Those secondary failures look like duplicate-render bugs and are not.

## Two details that matter if this is ever touched

- **Map-backed `Storage`, not no-op stubs.** These tests SET a value and assert the component reads it back ("stays dismissed across a remount", "scopes dismissal per tenant"). Stubs that dropped writes would convert 8 hard failures into 8 quiet false passes — strictly worse than the state being replaced.
- **Guarded on the VALUE, not the key.** `??=` and `"localStorage" in globalThis` both see Node's accessor and skip, leaving the bug in place. `!globalThis.localStorage` is what distinguishes it.

Ported from `tesserix-home#627`, which fixed the identical thing in the console.

## Scope

**Not** what #857 is about — CI runs Node 22, so it is unaffected either way, and this changes nothing there. It exists so the suite is runnable on 26. It sits beside the Pointer Capture, `scrollIntoView` and `ResizeObserver` stubs already in `vitest.setup.ts` for the same class of environment gap.

## Verification

`MobileAppPrompt.test.tsx` passes on v22.19.0, v24.20.0 and v26.5.0. Full `apps/admin` suite green.

Note: CI on this branch will fail the **npm audit gate** on advisories published today (`next` critical, `@tiptap/core`, `js-yaml`, `sharp`). That is repo-wide and unrelated — this PR touches one test-setup file and no dependency file. See the issue filed alongside this for why those cannot currently be patched.

### DIFF
--- a/apps/admin/vitest.setup.ts
+++ b/apps/admin/vitest.setup.ts
@@ -5,3 +5,57 @@ import { cleanup } from "@testing-library/react";
 afterEach(() => {
   cleanup();
 });
+
+// Web Storage, which this environment supplies only on some Node versions.
+//
+// `window.localStorage` reads as `undefined` under Node 26 even though the
+// document has a real origin and jsdom implements Storage:
+//
+//   - Node 26 defines `globalThis.localStorage` itself, as an accessor that
+//     returns `undefined` unless the process was started with
+//     `--localstorage-file` (it warns exactly that).
+//   - vitest's jsdom environment populates globals only for keys not already
+//     present. The key IS present — as Node's always-undefined accessor — so
+//     jsdom's real Storage never lands on it.
+//
+// Measured, because the boundary is narrower than "new Node":
+//
+//   v22.19.0  absent              MobileAppPrompt passes
+//   v24.20.0  absent              MobileAppPrompt passes
+//   v26.5.0   present, undefined  MobileAppPrompt fails 8 tests
+//
+// So this is NOT what #857 is about — CI runs 22 today and is unaffected
+// either way. It is here so the suite is runnable on 26, where
+// `MobileAppPrompt.test.tsx` otherwise dies in its `afterEach` on
+// `localStorage.clear()` and takes `cleanup()` with it, leaving later tests
+// to fail with "Found multiple elements" — failures that look like duplicate
+// renders and are not.
+//
+// Ported from tesserix-home#627, which fixed the identical thing in the
+// console.
+//
+// A real Map-backed Storage rather than no-op stubs: these tests SET a value
+// and assert the component reads it back ("stays dismissed across a remount",
+// "scopes dismissal per tenant"), so stubs that dropped writes would convert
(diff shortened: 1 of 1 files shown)


mahesh-sangawar:
Rebased onto current main and rebuilt to just its net change.

**Why it needed touching.** The green run on this PR was from 01:29Z, before #863 moved every gate from Node 22 to **24** — so those checks never ran on the gates that exist now. Merging on them would have been merging on stale evidence.

A straight rebase conflicted twice: the branch's history still carries the `appCredentials` commit that #863 has since landed with a better diagnosis. Since the PR's net diff against main was already only `apps/admin/vitest.setup.ts` (+54), the branch is rebuilt from main with exactly that and nothing else. No behaviour change from what was reviewed.

**Re-verified across all three Node versions**, from `npm ci` with npm@10.9.3:

| | MobileAppPrompt + appCredentials |
|---|---|
| v22.19.0 | 16 passed |
| v24.20.0 | 16 passed |
| v26.5.0 | 16 passed |

On 26 the run prints `ExperimentalWarning: localStorage is not available because --localstorage-file was not provided`, which is the accessor this PR describes, observed rather than inferred.

**Control run**, because a passing test proves nothing about what is making it pass: with the `vitest.setup.ts` change stashed, `MobileAppPrompt.test.tsx` fails **8/8** on v26.5.0 and passes with it restored. The fix is load-bearing.

Two lines of the description are now stale and worth ignoring: CI runs Node 24, not 22, and the npm audit gate note is resolved — #867 and #868 patched those advisories.

PR #866, opened 2026-09-09
fix(deps): patch the Next.js image-optimization RCE and three other high advisories

mahesh-sangawar:
Four high/critical advisories were published on 2026-09-08 between 20:51Z and 21:25Z. They block **every PR in the repo** — the identical audit gate passed on `e830e33` at 17:08Z and failed on the same commit plus one markdown file at 23:56Z. One of them is reachable in production.

This is the same failure mode the `audit_allowlist` comment in `ci.yml` already documents from 2026-09-02: "main went red on an unchanged lockfile when npm's feed picked them up". Unlike that one, these all have a version to point at.

## The one that matters: GHSA-2xp9-vwfh-vxw4 (critical)

Unauthenticated RCE in the Next.js Image Optimization API, via `libheif` in `sharp` when an **AVIF source** is optimized. Affects `next >= 16.0.0, < 16.3.3`; we were on **16.2.11**.

It is reachable on the public storefront:

- `/_next/image` is live and optimizing on `mark8ly.com` — no-args returns 400, a legitimate local image returns 200, unauthenticated. (Endpoint reachability only; no exploitation was attempted.)
- `apps/storefront/next.config.ts` allows `storage.googleapis.com` and `*.storage.googleapis.com` in `remotePatterns`. **Anyone can create a public GCS bucket**, so for this advisory the remote-source allowlist is not a meaningful gate — an attacker-hosted AVIF is one request away.

`apps/admin` carries the same GCS patterns. `apps/onboarding` declares no `images` config.

## The other three

| Advisory | Package | Was | Now |
|---|---|---|---|
| GHSA-p293-qw3h-jr36 (critical) | `next` | 16.2.11 | 16.3.4 |
| GHSA-rgj7-g3m4-5g8c (high) | `sharp` | 0.35.3 | 0.35.4 |
| GHSA-2883-xcg3-v3hh (high) | `js-yaml` | 4.3.1 / 3.15.1 | 4.3.2 / 3.15.2 |
| GHSA-j95f-988m-3j2f (high) | `@tiptap/core` | 3.22.5 | 3.31.3 |

**GHSA-p293-qw3h-jr36 does not apply to us** — it is RCE on *windows-hosted* servers and every deployed image is Alpine. It is patched by the same bump, so it is fixed incidentally rather than urgently.

## One change CI did not ask for

`fast-uri` 3.1.5 → 3.1.7 (GHSA-5jgf-p345-68v8 and three siblings, high, published 2026-09-02). The gate did **not** flag it, because `npm audit --workspace=` did not surface it for the five audited workspaces — I found it running the gate locally over the whole tree.

Included because the root `overrides` block already pins `fast-uri` for an earlier advisory in this exact family, so the fix is raising a floor that is already there. Flagging it explicitly since it widens the diff beyond what CI demanded.

A new `js-yaml@^3.13.1` override entry was needed too: the 3.x transitives under `cosmiconfig` and `@istanbuljs/load-nyc-config` are a separate specifier from the two 4.x pins already present, so the existing entries could not reach them.

## Verification

Run on **Node 22**, which is what `nextjs-ci.yml` uses (`node_version: "22"`).

- **The gate itself, reproduced locally** with the workflow's exact allowlist (`GHSA-w3rx-r6r6-pgpr`, `GHSA-5p2g-fcmc-qvqq`): **no blocking high/critical advisories**. The seven `image-size`/metro/react-native packages remain tolerated, exactly as before.
- `npm run test`: **1614 tests pass** — admin 1269, storefront 250, mobile-shared 95.
- `npm run check-types`: 5/5 clean. `npm run lint`: 4/4, 0 errors.
- `npm run build`: all three apps compile on Next 16.3.4.

### One pre-existing failure worth naming

On **Node 24**, `tests/unit/api/subscription/appCredentials.test.ts` fails 2 tests — Node 24's `FormData` converts a `Blob` to a `File` when a filename is given. It passes on Node 22, it is unrelated to this bump, and it is what #857 / #863 are about. I checked rather than assumed: the same two tests pass on 22 and fail on 24 with this branch checked out.

## Note on `next-env.d.ts`

The three `next-env.d.ts` files gain `import "./.next/types/root-params.d.ts"`. Generated by the 16.3 build, not hand-edited.


mahesh-sangawar:
---

## Update — two install-shape problems, found and fixed before merge

The first push went red twice, and neither failure was in the dependency versions. Both were the lockfile's **shape**, and both are worth recording because the second one is invisible until a clean install.

**1. I regenerated the lockfile with the wrong npm.** `packageManager` pins `npm@10.9.3`; my Node 24 shell runs npm 11.18. That rewrote hoisting across the tree. Redone with `npx npm@10.9.3`, which is what CI uses.

**2. `next` silently de-hoisted, and that broke React.** All three apps moved to the same `^16.3.3` and resolved the same `16.3.4`, but npm placed three separate copies under `apps/*/node_modules/next` instead of one at the root. `apps/admin` then loaded `next` from its own tree while `react-dom` still came from the root — two React instances, so every `useContext` returned null and **66 admin tests failed** with `Cannot read properties of null (reading 'useContext')`.

`npm dedupe` restored a single root `next@16.3.4`. Verified by resolution rather than by assertion:

```
base:  next -> <root>/node_modules/next
before dedupe: next -> apps/admin/node_modules/next   <- the bug
after dedupe:  next -> <root>/node_modules/next
```

**3. A latent bug the cleanup exposed.** `packages/mobile-shared` imports `expo-haptics` in `haptics/feedback.ts` but declares it only as a `peerDependency: "*"` — and with `legacy_peer_dependencies: true`, peers are not installed. It compiled only because the old lockfile carried a **stale root `expo-haptics@14.0.1`** that nothing declares (the sole declarer is `apps/mobile-admin` at `~56.0.0`, which 14.0.1 does not satisfy). Re-resolving removed the orphan and `check-types` immediately failed.

Fixed by declaring it where it is used — `expo-haptics: ~56.0.0` in `mobile-shared`'s `devDependencies` — rather than by preserving an orphaned lockfile entry.

## Re-verification, all from `npm ci` with npm@10.9.3 on Node 22

- Audit gate with the workflow's allowlist: **no blocking high/critical**.
- `test` + `check-types` + `lint`: **13/13 tasks, 1614 tests pass** (admin 1269, storefront 250, mobile-shared 95).
- `build`: all three apps compile.

A control run pinned the diagnosis rather than assuming it: the same clean install on the base commit passes `SignInForm.test.tsx` 18/18, so the failure was genuinely mine and is genuinely gone.

mahesh-sangawar:
Superseded by #867, which landed the same version bumps (next 16.3.4, sharp 0.35.4, js-yaml 4.3.2/3.15.2, fast-uri 3.1.7, @tiptap/core 3.31.3) while this branch was in verification. Closing rather than rebasing — the lockfile half is done.

Two things from this branch are NOT on main and I'll raise them separately rather than leave them in a closed PR:

1. **The three apps still declare `"next": "^16.2.11"`.** The lockfile pins 16.3.4 so `npm ci` is safe today, but the declared floor still admits the vulnerable range — the security fix lives only in the resolved lockfile, not in what the apps ask for. Same for the `sharp` override, still `>=0.35.0` rather than `>=0.35.4`.

2. **`packages/mobile-shared` imports `expo-haptics` but declares it only as `peerDependency: "*"`**, which `--legacy-peer-deps` never installs. It resolves today only because a stale root `expo-haptics@14.0.1` sits in the lockfile that nothing declares (the sole declarer is `apps/mobile-admin` at `~56.0.0`, which 14.0.1 does not satisfy). Still present on main. The next lockfile regeneration that removes the orphan will fail `@repo/mobile-shared#check-types` — which is exactly what happened on this branch.

Also worth recording for whoever regenerates the lockfile next: `packageManager` pins npm@10.9.3, and nvm ships npm 11.18 on both Node 22 and 24 here. Regenerating with the wrong npm de-hoisted `next` into all three apps, giving `apps/admin` its own React while `react-dom` stayed at the root — 66 admin tests failed with `Cannot read properties of null (reading 'useContext')`. `npx npm@10.9.3 dedupe` fixes it; verify with `require.resolve('next/package.json', {paths:['apps/admin']})` pointing at the repo root, not the app.

PR #868, opened 2026-09-09
fix(deps): declare the patched next and sharp floors so the fix survives a re-resolve

mahesh-sangawar:
Follow-up to #867, which patched the advisories in the **lockfile**. Three things it left behind, found while verifying the same fix on a parallel branch (#866, now closed as superseded).

## 1. The apps still ask for a vulnerable version

All three still declare `"next": "^16.2.11"`, and the root override still says `"sharp": ">=0.35.0"`.

`npm ci` is safe today because the lockfile pins 16.3.4 / 0.35.4. But the *declared floor* still admits the range the critical RCE (GHSA-2xp9-vwfh-vxw4, `>= 16.0.0 < 16.3.3`) lives in — the security fix exists only as a resolved artifact, not as a stated requirement. Anything that re-resolves against the declared ranges is free to land back inside the vulnerable window.

Raised to `^16.3.3` and `>=0.35.4`: the patched version becomes the floor, so the fix survives the next lockfile regeneration rather than depending on it.

## 2. `mobile-shared` imports a package it does not declare

`packages/mobile-shared/haptics/feedback.ts` has `import * as Haptics from "expo-haptics"`, but the package declares it only as `peerDependency: "*"` — and with `legacy_peer_dependencies: true`, peers are never installed.

It resolves today purely by accident: the lockfile carries a root `expo-haptics@14.0.1` that **nothing declares**. The only declarer is `apps/mobile-admin` at `~56.0.0`, which 14.0.1 does not satisfy — it is an orphan from an older tree.

Any lockfile regeneration that tidies the orphan breaks `@repo/mobile-shared#check-types` with `Cannot find module 'expo-haptics'`. That is not hypothetical: it happened on #866 the moment the tree was re-resolved. Fixed by declaring it in `devDependencies` (`~56.0.0`, matching mobile-admin) rather than by preserving an orphan. Root `expo-haptics` is now 56.0.3 — a version something actually asks for.

## 3. A trap for whoever regenerates the lockfile next

`packageManager` pins **npm@10.9.3**, and nvm here ships npm 11.18 on both Node 22 and Node 24. Regenerating with the wrong npm rewrites hoisting.

On #866 that put a separate `next` copy under each of `apps/*/node_modules`, so `apps/admin` loaded `next` from its own tree while `react-dom` still came from the root — two React instances, every `useContext` null, and **66 admin tests failing** with `Cannot read properties of null (reading 'useContext')`. Nothing in the diff looked wrong; the symptom points nowhere near the cause.

`npx npm@10.9.3 dedupe --legacy-peer-deps` is the fix, and the check is resolution rather than the lockfile's text:

```
$ node -e "require.resolve('next/package.json',{paths:['apps/admin']})"
<root>/node_modules/next/package.json     ← correct
apps/admin/node_modules/next/package.json ← the bug
```

## Verification

Everything from `npm ci` with **npm@10.9.3 on Node 22**, matching `nextjs-ci.yml`.

- **Hoisting checked by resolution**, not by reading the lock: all three apps resolve `next` to `<root>/node_modules/next`, and the lockfile holds exactly one copy at 16.3.4.
- `test` + `check-types` + `lint` + `build`: **16/16 tasks**, **1614 tests pass** (admin 1269, storefront 250, mobile-shared 95), all three apps compile.

The lockfile diff is large because `dedupe` rewrites placement, not versions. The versions that matter are unchanged from #867: `next` 16.3.4, `sharp` 0.35.4, `js-yaml` 4.3.2/3.15.2, `fast-uri` 3.1.7, `@tiptap/core` 3.31.3.

### DIFF
--- a/apps/admin/next-env.d.ts
+++ b/apps/admin/next-env.d.ts
@@ -1,6 +1,7 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
 import "./.next/types/routes.d.ts";
+import "./.next/types/root-params.d.ts";
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
--- a/apps/admin/package.json
+++ b/apps/admin/package.json
@@ -43,7 +43,7 @@
     "isomorphic-dompurify": "^3.22.0",
     "lucide-react": "^1.8.0",
     "marked": "^18.0.0",
-    "next": "^16.2.11",
+    "next": "^16.3.3",
     "papaparse": "^5.5.3",
     "postcss": "^8.5.8",
     "react": "19.2.8",
--- a/apps/onboarding/next-env.d.ts
+++ b/apps/onboarding/next-env.d.ts
@@ -1,6 +1,7 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
 import "./.next/types/routes.d.ts";
+import "./.next/types/root-params.d.ts";
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
--- a/apps/onboarding/package.json
+++ b/apps/onboarding/package.json
@@ -20,7 +20,7 @@
     "class-variance-authority": "^0.7.1",
     "clsx": "^2.1.1",
     "framer-motion": "^13.1.0",
-    "next": "^16.2.11",
+    "next": "^16.3.3",
     "postcss": "^8.5.8",
     "react": "19.2.5",
     "react-dom": "19.2.5",
--- a/apps/storefront/next-env.d.ts
+++ b/apps/storefront/next-env.d.ts
@@ -1,6 +1,7 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
 import "./.next/types/routes.d.ts";
+import "./.next/types/root-params.d.ts";
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
--- a/apps/storefront/package.json
+++ b/apps/storefront/package.json
@@ -20,7 +20,7 @@
     "@tailwindcss/postcss": "^4.2.2",
     "@tesserix/otto-widget": "^0.6.0",
     "@tesserix/web": "^2.4.2",
-    "next": "^16.2.11",
+    "next": "^16.3.3",
     "postcss": "^8.5.8",
     "react": "19.2.7",
     "react-dom": "19.2.7",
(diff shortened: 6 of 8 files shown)


mahesh-sangawar:
## Rescoped — main moved twice while this was in verification

#867 and #863 both landed since this PR was opened, and between them they covered two of the three things it originally carried. Rebased onto current main and reduced to what is genuinely still missing.

**Dropped — #863 already did it.** `packages/mobile-shared` now declares its expo modules (`fix(mobile-shared): declare the expo modules its type-check needs instead of borrowing a sibling app's hoist`), pinning `expo-haptics` at `~14.0.1`. That is a different choice from the `~56.0.0` this PR proposed, and it is the one on main — keeping mine would have regressed it for no benefit. The latent-orphan problem is closed either way.

**Still missing, and this is now the whole PR.** On current main:

```
apps/{admin,storefront,onboarding}  "next": "^16.2.11"
root overrides                      "sharp": ">=0.35.0"
```

Both still admit the versions the advisories name — `next` GHSA-2xp9-vwfh-vxw4 covers `>= 16.0.0 < 16.3.3`, and `sharp` GHSA-rgj7-g3m4-5g8c covers `< 0.35.4`. `npm ci` is safe today only because the lockfile happens to pin 16.3.4 and 0.35.4. The fix exists as a resolved artifact and nowhere in what the packages actually ask for, so any re-resolve against the declared ranges is free to land back inside the vulnerable window.

Raised to `^16.3.3` and `>=0.35.4`, so the patched version is the floor rather than a coincidence.

## Re-verified on Node 24

#863 moved every gate from Node 22 to 24, which invalidated this branch's original Node 22 verification — so it was re-run rather than assumed. From `npm ci` with **npm@10.9.3 on Node 24**:

- `test` + `check-types` + `lint` + `build`: **16/16 tasks**, **1614 tests pass** (admin 1269, storefront 250, mobile-shared 95), all three apps compile. That includes the `appCredentials` suite #863 repaired for Node 24.
- **Hoisting verified by resolution**, not by reading the lockfile — all three apps resolve `next` to `<root>/node_modules/next`, and the lock holds exactly one copy at 16.3.4:

```
$ node -e "require.resolve('next/package.json',{paths:['apps/admin']})"
<root>/node_modules/next/package.json
```

That check is here because getting it wrong is silent: an earlier attempt de-hoisted `next` into each app, which gave `apps/admin` its own React while `react-dom` stayed at the root, and 66 admin tests failed on `useContext` being null with nothing in the diff looking wrong. `npx npm@10.9.3 dedupe --legacy-peer-deps` is the fix — and note `packageManager` pins npm@10.9.3 while nvm ships 11.18 on both Node 22 and 24, so the pinned npm has to be forced.

The lockfile diff is large because `dedupe` rewrites placement. No dependency version changes from main.

Issue #858, opened 2026-09-08
34 of 35 e2e specs still run nowhere — they need a backend stack CI does not stand up


mahesh-sangawar:
Split out of #834, which is otherwise delivered. #851 triaged the suite and #855 added the route manifest; this is the part those deliberately left alone.

## Where the suite stands after #851

```
apps/admin/tests/e2e/*.spec.ts         26
apps/onboarding/tests/e2e/*.spec.ts     7
apps/storefront/tests/e2e/*.spec.ts     2
                                    -----
                                       35   specs presented as the enforced tier
apps/admin/tests/operator/*.spec.ts     6   opt-in, correctly labelled, not in scope here
```

**One of the 35 runs in CI** — `apps/admin/tests/e2e/pricing.spec.ts`, via `.github/workflows/e2e-runnable.yml`, behind an exact pass-count canary so a suite that silently runs nothing cannot report green. That canary is the right shape and should stay.

The other 34 sit under `tests/e2e/` looking enforced and are executed by nothing.

## Why the remaining 34 are not simply "wire them up"

They need a backend, and one of them shows the trap clearly. `apps/onboarding/tests/e2e/{form-validation,tax-id-migration}.spec.ts` assert only client-side state, so they read as backend-free — and they still time out. `apps/onboarding/app/onboarding/page.tsx` is `export const dynamic = "force-dynamic"` and imports `locations` from `@/lib/api/platform-api`, fetching reference data at render time with no fallback:

```
apps/onboarding/app/onboarding/page.tsx:3   import { locations } from "@/lib/api/platform-api";
apps/onboarding/app/onboarding/page.tsx:47  // Server component: fetches reference data once on the server
```

**A spec can assert only client state and still need a backend, because the page will not render without one.** Auditing specs by reading their assertions will therefore mis-classify them; #834's runnable subset was established by running each candidate, not by reading it, and that method should carry forward.

## The options, and the cost of each

1. **Stand up a real stack in CI** — marketplace-api + platform-api + Postgres. Highest signal, and the only option that addresses #834's second problem (mocked e2e cannot catch contract drift) for the flows the route manifest does not cover. Also the most expensive, and the shared `db-f1-micro` budget constraint is a real factor.
2. **Give the server components a fallback** so pages render without a backend, widening the backend-free subset. Cheaper, but it changes product code to suit tests, and a fallback that silently renders an empty country list is its own hazard.
[...]

PR #880, opened 2026-09-10
Unblock CI: grpc CVE-2026-84445, plus two defects in the e2e gates I added

mahesh-sangawar:
Unblocks every PR in the repo. Not related to any feature work.

## What broke

A HIGH advisory published against `google.golang.org/grpc` v1.83.1 after `62087078` merged. Trivy's **Scan image** step now fails on the Go service images:

```
│ google.golang.org/grpc │ CVE-2026-84445 │ HIGH │ fixed │ v1.83.1 │ 1.82.2, 1.83.2, … │
```

`main` was green at 18:37 today and nothing in the repo changed — this is the "a new advisory turns every PR red" pattern, and it will keep doing so until grpc moves.

## The fix

`go get google.golang.org/grpc@v1.83.2` + `go mod tidy` in the four services that carry it:

| service | before | after |
|---|---|---|
| `auth-bff` | v1.83.1 (indirect) | v1.83.2 |
| `platform-api` | v1.83.1 (indirect) | v1.83.2 |
| `marketplace-api` | v1.83.1 (direct) | v1.83.2 |
| `otto` | v1.83.1 (indirect) | v1.83.2 |

v1.83.2 is the lowest listed fixed version, so this is a patch bump rather than a major move.

## Verification

`go build ./...` and `go vet ./...` clean in all four. CI runs the full Go suites and re-scans the images, which is the real check — the point of this PR is that the scan goes green.

Diagnosed by step, not by job name: both failing jobs failed at **Scan image**, with Build and Push succeeding.


mahesh-sangawar:
## Scope widened — two defects in the e2e gates I added in #878

The CVE fix alone could not merge: the `admin suite vs real stack` job kept failing it, and that job is mine. Both defects are in gates I wrote, and both are the kind that teach people to re-run rather than to look.

**1. The health gate was flaky by construction** (`8666bee9`). It polled platform-api in a loop, then checked auth-bff, marketplace-api and Zitadel **immediately, with no wait** — so whichever service came up slowest decided the outcome. It passed on the PR that introduced it and failed on the very next unrelated PR with `marketplace-api never became healthy`. Now every service gets its own wait, and a timeout dumps compose logs.

**2. Flakiness was reported as a broken stack** (`5d1307b1`). Playwright counts a retry-then-pass as `flaky`, not `passed`, so an exact-`11 passed` gate fires with *"the suite is not actually executing and passing"* — which sends the reader hunting for a dead service. `tax-id-migration.spec.ts:87` (`shows US-specific help text`) did exactly this on this PR: `10 passed, 1 flaky, 5 skipped`.

Flaky stays a **failure** — it is a real defect and tolerating it is how a suite rots — but it now says so in its own message, and tells you to fix the intermittency rather than re-run.

That intermittency is real and still open; it wants its own investigation, tracked on #858.

Issue #696, opened 2026-09-05
marketplace-api's Stripe billing key points at an unpopulated test surface, so every subscribe fails at price resolution


mahesh-sangawar:
marketplace-api's billing key points at a Stripe surface nobody publishes prices to, so every subscribe fails at price resolution.

Found while investigating #650. Unrelated to that issue's subject, and larger than it.

## The measurement

Four Stripe surfaces are in play. Only one of them is the one marketplace-api uses, and it is the only one with no prices in it.

| account | mode | active `mark8ly_*` prices | who points here |
|---|---|---|---|
| `acct_1SgwbFCyiazmanuP` (Tesserix main) | **live** | **42** | console live read/write keys |
| `acct_1SgwbhE1SGxvhzVd` (sandbox) | test | **42** | console test read key |
| `acct_1SgwbFCyiazmanuP` (Tesserix main) | **test** | **0** | ← `mark8ly-stripe-billing` / `STRIPE_BILLING_SECRET_KEY` |
| `acct_1U8wRZE0e9LEcqjJ` | test | — | `dev-tesserix-stripe-rehearsal-write-key` only |

The decisive probe is the exact lookup key the code resolves:

```
GET /v1/prices?lookup_keys[]=mark8ly_pro_monthly_developed_v1
  console sandbox (acct_1Sgwbh…) -> 1 match
  main account LIVE              -> 1 match
  main account TEST  (ours)      -> 0 matches
```

The main account's test mode returns `{"data": []}` for products, prices, customers, subscriptions, invoices and coupons alike — successful reads of an empty surface, not permission denials (the same key is denied on `/v1/account`, which is what a denial looks like).

## Why this breaks billing outright

`billingstripe.PriceIDFor` resolves every plan price through `FindPriceByLookupKey`, which returns `ErrNotFound` on an empty list:

```go
// services/marketplace-api/internal/billing/stripe/price.go:22-35
for p, err := range c.sdk.V1Prices.List(ctx, listParams) { ... return mapPrice(p), nil }
return nil, ErrNotFound
```

With the currently-mounted key that branch is unreachable for **every** plan/period/currency/tier combination. So `trial.CreateSubscription` and both `planchange` callers fail before they reach Stripe's subscription API.

## This reframes the "0 rows" reasoning in #371 and #366

[...]

PR #882, opened 2026-09-21
fix(billing): gate subscribe and cancel to 503, and correct the AU tax + go-live docs

mahesh-sangawar:
Two related changes, both prompted by a launch-readiness assessment that turned up something not previously recorded.

## The finding

**Production has run a live Stripe key since 2026-09-09** (`tesserix-k8s` `6b8f91b7`, #366/#371 closed). `docs/runbooks/stripe-billing-go-live.md` still said *"not yet — production stays on the test key"* and was never updated.

Verified against the running cluster on 2026-09-21: the admin pod consumes `prod-mark8ly-stripe-billing-secret-key-live` (prefix `rk_live_`), `CONSOLE_CATALOG_MODE=live`, account `acct_1SgwbFCyiazmanuP` with 42 active live prices.

**Nothing has been charged.** 0 live customers, subscriptions, invoices, charges and refunds (read-only Stripe `GET`s, all 200-with-empty rather than 403); 0 non-empty Stripe ids in `store_subscriptions`. The exposure is entirely latent — for now.

The swap itself was done correctly: key, webhook secret and catalog mode all moved in one commit, which is the atomicity the runbook asked for. The problem is what it made reachable.

## Why this is urgent rather than merely untidy

**Cancellation never reaches Stripe.** There is no `Subscriptions.Cancel` call anywhere in marketplace-api, and `UpdateSubscriptionParams` has no cancel field — `cancel_at_period_end` is only ever *read from* Stripe payloads. The in-code justification is circular: `internal/subscription/cancel/service.go:68-70` says Stripe "already has `cancel_at_period_end=true` from the webhook layer", but the webhook layer only mirrors Stripe → local.

So for a real subscriber:
- the UI tells them *"full access continues until {periodEnd}"* (`apps/admin/lib/copy/subscription.ts:107-109`) — false;
- `current_period_end` is written only from processed webhooks and never at subscribe time, so `FinalizeCron` sees `NULL` and expires the row, revoking access at the next tick;
- Stripe, knowing nothing of the cancellation, keeps billing.

The swap is also what made price resolution work for the first time (the previous key sat on the main account's empty test mode, #696), and the billing routes are wired with no feature flag. **The next merchant to click Subscribe is simultaneously the first real charge and the first to hit this bug.**

## What this PR does

**1. Gates the two routes that can create live money** (`fix(billing)`)

`RequireBillingWrites` middleware on `POST /billing/subscription` and `POST /subscription/cancel`, returning 503 `billing_writes_disabled` with merchant-safe copy. Fail-closed: `BILLING_WRITES_ENABLED` defaults to `false`, so an environment that forgets to set it blocks writes rather than taking money by accident.

Mounted *after* the authz middleware, so the gate can't be used to probe permissions. Enabled in `infra/dev/docker-compose.yml` so local development against the sandbox key is unaffected.

Not gated: `POST /subscription/change-plan`, which also writes prorated invoices. It is inert today because there are zero live subscriptions for it to act on. Flagged in the runbook; say the word and I'll add it.

**2. Reverses four documents that instruct enabling Stripe Tax for AU** (`docs(billing)`)

`p20-prod-launch-readiness.md` (B1.2 and the billing checklist) and `subscription-model-design.md` (§19.4 and the effort table) all told a reader to switch Stripe Tax on for AU. Tesserix is **not GST-registered**, and AUD prices are `tax_behavior: exclusive` — so acting on those lines adds 10% on top silently, with no repricing and no copy change, collecting GST there is no registration to remit. The runbook names that exact outcome as *"the worse outcome"*.

Inverted rather than deleted, so the reasoning survives. §19.4's claim that "Mark8ly Pty Ltd charges 10% GST + remits to ATO" was never true of the live system and is now marked superseded. The `tax_behavior: exclusive` requirement is kept and called out as load-bearing — it is what keeps registration a config flip.

Also corrects the runbook's headline status, and records that the rollback target is the **`-sandbox`** pair, not `-test` — both hold `rk_test_` but they are different accounts, and `-test` is the empty one that caused #696.

## Testing

`RequireBillingWrites` is covered test-first both ways (blocks when disabled with the right code and a non-empty message; transparent when enabled). `go build ./...`, `go vet` and the package tests pass. `docker compose config` validates.

No e2e spec exercises either route — the only test touching them is a frontend unit test that mocks the API — so CI is unaffected.

## Not in this PR

The gate is a stopgap, not the fix. Before it comes off: cancellation must reach Stripe, `current_period_end` must be persisted at subscribe time, and the merchant-facing copy needs re-checking. The assessment also surfaced a missing startup key-mode assertion, a fail-open `CONSOLE_CATALOG_MODE`, and `reconciliation-cron` not being built into the image while a live alert depends on it.


mahesh-sangawar:
Superseded by #885. This PR was auto-closed by GitHub when its base branch (#883's) was deleted on merge — my mistake for passing `--delete-branch` while this one was still targeting it. A closed PR whose base branch no longer exists cannot be reopened or retargeted, so the same two commits were replayed onto `main` and opened as #885. No content change.

